PR #3, opened 2018-12-13
Prepare Travis, CodeClimate, and deployment

Fryguy:
Prepare Travis, CodeClimate, and deployment


Fryguy:
This is going to be red for the time being until we can have common set up its own database (as Authentication specs require an actual database).

I've already started pulling over the concepts as we've done in topological_inventory-core, however, I can't find the database migrations for Authentication and Encryption.

Fryguy:
@agrare Thoughts on making this look like topological_inventory-core? I feel like they are similar in that they will have shared "stuff", including some models.  Last commit is a WIP PR where I've started moving all of that around.

agrare:
@Fryguy most of that was so that we could run with just activerecord, if this is going to be a normal rails engine we probably don't need all of that complexity and could use a simple dummy app?

Fryguy:
> if this is going to be a normal rails engine we probably don't need all of that complexity and could use a simple dummy app?

I'm not sure.  The only thing on the list so far that needs models is the auth stuff, which is the only thing in this repo at them moment.

> Currently I have a rails generator that will generate migrations for the using app e.g. SSP, Approval,( insert other microservices here ) but internally there are no migrations included with this gem.

I'm not sure I like the generator approach, mainly because as we add new migrations, then all dependent app need to regenerate the migrations, and we have to be sure it only pulls in the newest ones.  This feels complex.  If this is an engine, it can just bring the migrations directly.

Fryguy:
@syncrou This PR now just gets the basic badges and stuff in place, but it will be red because the tests are not passing, so it can be merged on red.  To get the tests working, instead this needs to be updated with a dummy app, which @bdunne started in #4.

PR #4, opened 2018-12-17
Get tests working

bdunne:
- Namespace in ManageIQ::API::Common not to conflict with existing ManageIQ::API namespace
- Follow the pattern for Rails engines
- Get tests running


Fryguy:
I have already opened #3 to get some of this working...can we just work in the same PR instead of opening new PRs?  There is overlap now and I can't wrap my head around it.

Fryguy:
@bdunne I killed my change in #3 because I was going down the "ActiveRecord-only" path, which was wrong...better to just be a normal Rails engine.

Are most of the files in this PR just from doing `rails new` ?

app/models/manageiq/api/common/application_record.rb
```diff
@@ -0,0 +1,9 @@
+module ManageIQ
+  module API
+    module Common
+      class ApplicationRecord < ActiveRecord::Base
```
Fryguy: Do we need a Common::ApplicationRecord - feels weird...I would just expect mixins.  Actually this is going to get weird in general if models are in 2 places, because you can't have 2 ApplicationRecords.  cc @syncrou

spec/dummy/.ruby-version
```diff
@@ -0,0 +1 @@
+ruby-2.4.3
```
Fryguy: Let's not commit a `.ruby-version`

spec/dummy/bin/setup
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env ruby
```
Fryguy: This isn't part of a vanilla app...do we need this?

spec/dummy/bin/update
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env ruby
```
Fryguy: Same comment

spec/models/authentication_spec.rb
```diff
@@ -1,22 +1,22 @@
...
-    end
-  end
-end
+#     it "returns the unencrypted version to the caller" do
+#       auth.reload
+#       expect(auth.secret).to eq "blah"
+#       expect(auth.secret).not_to match(/v2/)
+#     end
+#   end
+# end
```
Fryguy: I'm assuming this is temporary until we can figure out how to have testable migrations?

.gitignore
```diff
@@ -3,3 +3,4 @@
 /v2_key
 /.ruby-version
 /.ruby-gemset
+spec/dummy/log/
```
Fryguy: Should we just ignore all of spec/dummy?

Gemfile
```diff
@@ -1,11 +1,13 @@
-source "https://rubygems.org"
+source 'https://rubygems.org'
+git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
```
Fryguy: Let's just kill the git_source line

config/routes.rb
```diff
@@ -0,0 +1,2 @@
+ManageIQ::API::Common::Engine.routes.draw do
+end
```
Fryguy: I think we can remove routes.

lib/manageiq/api/common/engine.rb
```diff
@@ -0,0 +1,9 @@
+module ManageIQ
```
Fryguy: I think we need an inflection defined for ManageIQ as well as an initializer for it in the engine file.

spec/dummy/bin/setup
Fryguy: Oh huh...I guess it is.

spec/dummy/config/routes.rb
```diff
@@ -0,0 +1,3 @@
+Rails.application.routes.draw do
```
syncrou: Just noting this comment in reference to needing routes or not ( for the dummy app as well. ) https://github.com/ManageIQ/manageiq-api-common/pull/4/files#r243638109

spec/models/authentication_spec.rb
bdunne: Correct

Fryguy: My PR got this working, and I think it's pretty straightforward, but I'm ok merging as is, and then fixing this in a follow up PR.

PR #5, opened 2018-12-17
Paginated response

bdunne:


lib/manageiq/api/common/paginated_response.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  def link_with_new_offset(offset)
+    request_uri.dup.tap do |u|
+      u.query = query_hash_merge("offset" => offset.to_s)
+    end.to_s
+  end
+
+  def query_hash_merge(new_hash)
+    parsed_query.merge(new_hash).collect { |k, v| "#{k}=#{v}" }.join("&")
```
Fryguy: Might need URI escaping on the k and v (unless they are already escaped coming in )

Fryguy: There's also a method that does this for you.. CGI.www_encoded_form or something like that (Also URI.query_string or something )

bdunne: `Hash#to_query` seemed simplest

bdunne:
@Fryguy Any other concerns?

lib/manageiq/api/common/paginated_response.rb
```diff
@@ -0,0 +1,94 @@
...
+        def parsed_query
+          @parsed_query ||= Rack::Utils.parse_query(request_uri.query)
+        end
+
+        def request_uri
+          @request_uri ||= URI.parse(@request.original_url)
+        end
+
+        def current_limit_multiplier
+          @current_limit_multiplier ||= offset / limit
```
Fryguy: This is an integer division with flooring...is that ok (e.g. 20 / 3 # => 6)

spec/lib/manageiq/api/common/paginated_response_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    let(:base_query) { double("AR:Clause", :count => count) }
+    let(:request) { double("Request", :original_url => "http://example.com/resource?param1=true&limit=#{limit}") }
+
+    def url_with_offset(offset)
+      "http://example.com/resource?limit=#{limit}&offset=#{offset}&param1=true"
+    end
+
+    context "with 6 records, limit 2" do
+      let(:count) { 6 }
+      let(:limit) { 2 }
```
Fryguy: might be better to pick numbers that don't align mathematically (for example, count being 7, and limit being 2 would fine better edge cases).

spec/lib/manageiq/api/common/paginated_response_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      end
+
+      it "third page" do
+        expect(described_class.new(base_query: base_query, request: request, limit: 2, offset: 4).send(:links_hash)).to eq(
+          "first" => first_url, "last" => last_url, "next" => nil, "prev" => url_with_offset(2)
+        )
+      end
+    end
+
+    context "with 31 records, limit 10" do
```
Fryguy: Oh...why?  Feels like you are testing the exact same stuff?

spec/lib/manageiq/api/common/paginated_response_spec.rb
```diff
@@ -0,0 +1,66 @@
+describe ManageIQ::API::Common::PaginatedResponse do
+  context "private links methods" do
+    let(:base_query) { double("AR:Clause", :count => count) }
+    let(:request) { double("Request", :original_url => "http://example.com/resource?param1=true&limit=#{limit}") }
+
+    def url_with_offset(offset)
+      "http://example.com/resource?limit=#{limit}&offset=#{offset}&param1=true"
+    end
+
+    context "with 6 records, limit 2" do
```
Fryguy: The name of the context shouldn't describe what's in it, but instead should describe what the purpose of the context is, IMO.

spec/lib/manageiq/api/common/paginated_response_spec.rb
bdunne: That's why I have (6 / 2) and (31 / 10), I wanted to test both cases.

Fryguy: Ah ok, then as per my comment below, the context titles should probably be "when the count and limit align" and "when the count and limit do not align", which makes their purpose clearer and explains 2 tests.

PR #22, opened 2019-01-10
Loosen dependency on rails

bdunne:
Each application that includes this should specify its specific version of rails if necessary.


syncrou:
@bdunne - This looks good to me, but Hakiri is saying otherwise.

bdunne:
Reversing the order fixed it. ¯\\\_(ツ)_/¯

PR #23, opened 2019-01-17
Add a Metrics class which will add the Prometheus middleware

carbonin:


manageiq-api-common.gemspec
```diff
@@ -16,6 +16,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "acts_as_tenant"
   spec.add_runtime_dependency "manageiq-password", "~>0.1"
   spec.add_runtime_dependency "pg", "> 0"
+  spec.add_runtime_dependency "prometheus-client", "~> 0.8.0"
```
bdunne: Should we visually separate this from the rest of the runtime dependencies and note which optional feature it's here for?

lib/manageiq/api/common/config_editor.rb
```diff
@@ -0,0 +1,21 @@
+module ManageIQ
+  module API
+    module Common
+      class ConfigEditor
```
bdunne: Shouold this be more specific and include "Metrics" in the name?

carbonin: I was intending to add other methods for things that alter the rails application config object. The alternative approach is to have a bunch of classes that look like this one, but if we do runtime requires for all of the different features it still remains "opt-in" this way and it doesn't look like there will be too many extra helpers and such.

manageiq-api-common.gemspec
carbonin: If that's something we're going to start doing then I'm fine with it, but isn't tenancy an optional feature also?

lib/manageiq/api/common/config_editor.rb
bdunne: Oh, I was thinking something more along the lines of calling something like this in the engine definition
```ruby
ManageIQ::API::Common::Logging.install
ManageIQ::API::Common::Metrics.install
```

manageiq-api-common.gemspec
bdunne: It's not opt-in at the moment, but it probably should be.  And I think that dependency should be visually separate also.

lib/manageiq/api/common/config_editor.rb
carbonin: Yeah, I'm fine with either. As I started writing it it seemed like overkill, but I can go that direction if you feel strongly about it. @Fryguy any opinions here?

PR #25, opened 2019-01-17
Maximums and minimums for limit and offset

bdunne:
Limit:
- Less than 1 doesn't make sense
- More than 1000 (guess) may cause server problems (timeouts / lots of memory used)

Offset:
- Negative numbers don't make sense

### DIFF
--- a/lib/manageiq/api/common/paginated_response.rb
+++ b/lib/manageiq/api/common/paginated_response.rb
@@ -7,7 +7,7 @@ module ManageIQ
         def initialize(base_query:, request:, limit: nil, offset: nil)
           @base_query = base_query
           @request    = request
-          @limit      = (limit || 100).to_i
+          @limit      = (limit || 100).to_i.clamp(1, 1000)
           @offset     = (offset || 0).to_i
         end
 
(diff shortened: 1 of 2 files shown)


lib/manageiq/api/common/paginated_response.rb
```diff
@@ -7,8 +7,8 @@ class PaginatedResponse
         def initialize(base_query:, request:, limit: nil, offset: nil)
           @base_query = base_query
           @request    = request
-          @limit      = (limit || 100).to_i
-          @offset     = (offset || 0).to_i
+          @limit      = (limit || 100).to_i.clamp(1, 1000)
+          @offset     = (offset || 0).to_i.clamp(0, Float::INFINITY)
```
Fryguy: you can leave off the second number entirely I think.

bdunne: Nope, https://github.com/ManageIQ/more_core_extensions/blob/master/lib/more_core_extensions/core_ext/numeric/clamp.rb#L9

Fryguy: Oh weird...

we should probably change clamp to follow the semantics for the new Ruby clamp method anyway, but even so, for now then, this is fine.

PR #28, opened 2019-01-22
Rename dummy database.yml to .sample

syncrou:
Ran into an issue where the dummy `database.yml` file was not configured to match my postgresql setup.

Renaming the file promotes a nicer error message that is easier to track:

```
RuntimeError:
  Cannot load database configuration:
  Could not load database configuration. No such file - ["config/database.yml"]
```

Also adds `.travis.yml` prebuild commands to copy the `.sample` file into `database.yml`


.travis.yml
```diff
@@ -14,6 +14,7 @@ before_install:
 before_script:
 - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
 - chmod +x ./cc-test-reporter
+- "cp spec/dummy/config/database.yml.sample spec/dummy/config/database.yml"
 - "./cc-test-reporter before-build"
```
Fryguy: These 3 statements at the top here are a unit for cc-test-reporter, so this new line should not interleave them

syncrou: Do I not need it at all? Or does this build not make use of `DATABASE_URL` and it is still needed?

Fryguy: yeah, you don't need it at all

Fryguy:
Let's name the file such that the .yml is at the end... `database.sample.yml`, for example, so that file parsers, syntax highlighting, etc works.

Additionally, add spec/dummy/config/database.yml to the .gitignore.

PR #29, opened 2019-01-23
Add common header utility

syncrou:
1. Sets / Get a thread safe version of any Rails header `ManageIQ::API::Common::Headers.current = headers` makes use of  `Thread.current[:attr_current_headers]`

2. Provides simple validation forcing a saved header using the Rails `ActionDispatch::Http::Headers` type.

Addresses this #29

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -1,6 +1,5 @@
 #!/usr/bin/env ruby
 
-require 'rails/all'
 require "bundler/setup"
 require "manageiq/api/common"
 
--- a/bin/console
+++ b/bin/console
@@ -1,5 +1,6 @@
 #!/usr/bin/env ruby
 
+require 'rails/all'
 require "bundler/setup"
 require "manageiq/api/common"
 
--- a/lib/manageiq/api/common.rb
+++ b/lib/manageiq/api/common.rb
@@ -1,8 +1,8 @@
 require "manageiq/api/common/engine"
+require "manageiq/api/common/headers"
 require "manageiq/api/common/inflections"
 require "manageiq/api/common/logging"
 require "manageiq/api/common/metrics"
-require "manageiq/api/common/headers"
 
 module ManageIQ
   module API
--- a/lib/manageiq/api/common.rb
+++ b/lib/manageiq/api/common.rb
@@ -2,6 +2,7 @@ require "manageiq/api/common/engine"
 require "manageiq/api/common/inflections"
 require "manageiq/api/common/logging"
 require "manageiq/api/common/metrics"
+require "manageiq/api/common/headers"
 
 module ManageIQ
   module API
--- a/lib/manageiq/api/common/headers.rb
+++ b/lib/manageiq/api/common/headers.rb
@@ -16,23 +16,26 @@ module ManageIQ
           HeaderScope.current
         end
 
-        def self.decode(key)
-          header = HeaderScope.current || nil
-          if header
-            val = header.instance_variable_get(:@req)
-            JSON.parse(Base64.decode64(val[key]))
-          end
+        def self.decode(headers, key)
+          validate_headers(headers)
+          val = headers.instance_variable_get(:@req).stringify_keys
+          JSON.parse(Base64.decode64(val[key]))
         end
 
         def self.encode(val, key)
-          Base64.strict_encode64(val[key].to_json)
+          if val.is_a?(Hash)
+            hashed = val.stringify_keys
+            Base64.strict_encode64(hashed[key].to_json)
+          else
+            raise StandardError, "Must be a Hash"
+          end
         end
 
         def self.validate_headers(val)
           if val.is_a?(ActionDispatch::Http::Headers)
-            return
+            true
           else
-            raise StandardError, 'Not a Header Class'
+            raise StandardError, 'Not an ActionDispatch::Http::Headers Class'
           end
         end
       end
--- a/lib/manageiq/api/common/headers.rb
+++ b/lib/manageiq/api/common/headers.rb
@@ -1,42 +1,19 @@
-module HeaderScope
-  thread_mattr_accessor :current
-end
-
 module ManageIQ
   module API
     module Common
       module Headers
-
         def self.current=(val)
           validate_headers(val)
-          HeaderScope.current = val
+          Thread.current[:attr_current_headers] = val
         end
 
         def self.current
-          HeaderScope.current
+          Thread.current[:attr_current_headers]
         end
 
-        def self.decode(headers, key)
-          validate_headers(headers)
-          val = headers.instance_variable_get(:@req).stringify_keys
-          JSON.parse(Base64.decode64(val[key]))
-        end
-
-        def self.encode(val, key)
-          if val.is_a?(Hash)
-            hashed = val.stringify_keys
-            Base64.strict_encode64(hashed[key].to_json)
-          else
-            raise StandardError, "Must be a Hash"
-          end
-        end
-
-        def self.validate_headers(val)
-          if val.is_a?(ActionDispatch::Http::Headers)
-            true
-          else
-            raise StandardError, 'Not an ActionDispatch::Http::Headers Class'
-          end
+        private_class_method def self.validate_headers(val)
+          raise ArgumentError, 'Not an ActionDispatch::Http::Headers Class' unless val.is_a?(ActionDispatch::Http::Headers)
+          true
         end
       end
     end
--- a/lib/manageiq/api/common/headers.rb
+++ b/lib/manageiq/api/common/headers.rb
@@ -1,0 +1,41 @@
+module HeaderScope
+  thread_mattr_accessor :current
+end
+
+module ManageIQ
+  module API
+    module Common
+      module Headers
+
+        def self.current=(val)
+          validate_headers(val)
+          HeaderScope.current = val
+        end
+
+        def self.current
+          HeaderScope.current
+        end
+
+        def self.decode(key)
+          header = HeaderScope.current || nil
+          if header
+            val = header.instance_variable_get(:@req)
+            JSON.parse(Base64.decode64(val[key]))
+          end
+        end
+
+        def self.encode(val, key)
+          Base64.strict_encode64(val[key].to_json)
+        end
+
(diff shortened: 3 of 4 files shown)


syncrou:
@Fryguy please review

cc - @mkanoor, @bzwei

lib/manageiq/api/common/headers.rb
```diff
@@ -0,0 +1,44 @@
...
+        end
+
+        def self.current
+          HeaderScope.current
+        end
+
+        def self.decode(headers, key)
+          validate_headers(headers)
+          val = headers.instance_variable_get(:@req).stringify_keys
+          JSON.parse(Base64.decode64(val[key]))
```
mkanoor: @syncrou Should there be a User class that can store the parsed properties and we could do 
User.current_user?

syncrou: @mkanoor - Yes, that will be set up as a separate PR.  I started with headers to make sure we were happy with the process etc.

mkanoor:
@syncrou How would we get access to the raw headers, would there be a get call to fetch it from the current thread, since we might want to pass those headers to other services after we have parsed it and created an in memory user object.

mkanoor:
@syncrou with managiq we used to have a with_user 
https://github.com/ManageIQ/manageiq/blob/8c32eeb70deb7f96d207059be48dce3b9e799f77/app/models/user.rb#L263

It basically does the same thing but allows you to pass a block to execute, so it would It be nice to wrap a controller action code inside the **with_request** function so we can clean up the thread local storage at the end.
It also helps with writing specs, otherwise you have thread variables pollution

https://github.com/ManageIQ/manageiq/blob/5f5b04b53e41bfd01f966fd881481309f688f070/spec/models/custom_button_spec.rb#L101

lib/manageiq/api/common/headers.rb
mkanoor: @syncrou 
From the headers how do you know which ones have to be base64 decoded and which ones have to be left alone.
If we get passed in a User-Agent in the HTTP header that data is not base64 encoded. Do we need to maintain a list of keys which are base64 encoded and which ones aren't.

lib/manageiq/api/common/headers.rb
```diff
@@ -0,0 +1,44 @@
...
+          if val.is_a?(Hash)
+            hashed = val.stringify_keys
+            Base64.strict_encode64(hashed[key].to_json)
+          else
+            raise StandardError, "Must be a Hash"
+          end
+        end
+
+        def self.validate_headers(val)
+          if val.is_a?(ActionDispatch::Http::Headers)
```
mkanoor: @syncrou If this is for x-rh-auth-identity should we check if he decoded value conforms to the IPP 

```
{
  "identity": {
    "account_number": "0369233",
    "type": "User",
    "user" : {
      "username": "jdoe",
      "email": "jdoe@acme.com",
      "first_name": "John",
      "last_name": "Doe",
      "is_active": true,
      "is_org_admin": false,
      "is_internal": false,
      "locale": "en_US"
    },
    "internal" : {
      "org_id": "3340851",
      "auth_type": "basic-auth",
      "auth_time": 6300
     }
  }
}

```

lib/manageiq/api/common/headers.rb
```diff
@@ -0,0 +1,44 @@
...
+        def self.current=(val)
+          validate_headers(val)
+          HeaderScope.current = val
+        end
+
+        def self.current
+          HeaderScope.current
+        end
+
+        def self.decode(headers, key)
```
mkanoor: @syncrou couldn't the headers be accessed from the HeaderScope.current? why does the caller have to pass it in?

lib/manageiq/api/common/headers.rb
syncrou: That's a good question.   I'm not sure if that is an implementation question for the microservice or something the common gem should handle.

lib/manageiq/api/common/headers.rb
syncrou: I'd like to discuss this more.  I originally was thinking ( and I think @Fryguy was too? ) we would have 2 separate modules.  One strictly for raw headers, and a `User` variant that validates the `x-rh-identity` values like what you're suggesting (among the other `.` notation attributes it would offer as well.) I do agree we need a mechanism to sanity check if the `x-rh-identity` key is valid.

lib/manageiq/api/common/headers.rb
syncrou: Yes if it was set before hand.  I'm open to either: not decoding if `ManageIQ::API::Common::Headers.current` is nil or as a stand alone utility with no dependencies.  I really wanted to have the conversation around the most obvious use cases to see what everyone thought would be best.

bin/console
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env ruby
 
+require 'rails/all'
```
Fryguy: This feels like it should not be necessary

lib/manageiq/api/common/headers.rb
```diff
@@ -0,0 +1,44 @@
+module HeaderScope
+  thread_mattr_accessor :current
```
Fryguy: This should be directly in Headers... Not sure why it's separated.

lib/manageiq/api/common/headers.rb
Fryguy: I agree with madhu here that it should just use current or be a private class method that can be treated in isolation.

lib/manageiq/api/common/headers.rb
```diff
@@ -0,0 +1,44 @@
...
+          val = headers.instance_variable_get(:@req).stringify_keys
+          JSON.parse(Base64.decode64(val[key]))
+        end
+
+        def self.encode(val, key)
+          if val.is_a?(Hash)
+            hashed = val.stringify_keys
+            Base64.strict_encode64(hashed[key].to_json)
+          else
+            raise StandardError, "Must be a Hash"
```
Fryguy: Should be ArgumentError

lib/manageiq/api/common/headers.rb
Fryguy: Yes a separate module should have the header processing.  Similar to madhu comment above, I don't think any Base64 decoding belongs here

bin/console
syncrou: I thought the same. Wasn't able to use the 'gem' console without it. Maybe there is another env or require at a higher level I'm missing.

lib/manageiq/api/common/headers.rb
syncrou: I wanted the validation to happen ahead of setting the accessor. Wasn't sure how else to declare the accessor and control validation ahead of setting it.

syncrou: @Fryguy do you think I shouldn't have any validation before setting `.current`? I guess I should ask that question too.

lib/manageiq/api/common/headers.rb
```diff
@@ -0,0 +1,44 @@
...
+end
+
+module ManageIQ
+  module API
+    module Common
+      module Headers
+
+        def self.current=(val)
+          validate_headers(val)
+          HeaderScope.current = val
```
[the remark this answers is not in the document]
bzwei: Why not directly call `Thread.current[:attr_current_headers] = val` after the validation?

syncrou:
@Fryguy - I applied the requested feedback.
One change, I pulled out `thread_mattr_accessor` use in `HeaderScope` and used `Thread.current` instead.  `thread_mattr_accessor` makes use of `Thread.current` as the base of it's operations: https://www.rubydoc.info/github/rails/rails/Module:thread_mattr_reader

That change allowed for minimal validation before the `ManageIQ::API::Common::Headers.current` is set.

All other `encode`, `decode` code was pulled out and will be moved to the more specific `User` module which will be built next.

syncrou:
@mkanoor, @bzwei - Please review

lib/manageiq/api/common.rb
```diff
@@ -2,6 +2,7 @@
 require "manageiq/api/common/inflections"
 require "manageiq/api/common/logging"
 require "manageiq/api/common/metrics"
+require "manageiq/api/common/headers"
```
Fryguy: minor, but alphabetize (unless there's a reason for the order?)

syncrou:
@Fryguy - Just pushed the alphabetical change as suggested.

lib/manageiq/api/common/headers.rb
```diff
@@ -0,0 +1,25 @@
...
+        def self.current
+          Thread.current[:attr_current_headers]
+        end
+
+        def self.validate_headers(val)
+          if val.is_a?(ActionDispatch::Http::Headers)
+            true
+          else
+            raise ArgumentError, 'Not an ActionDispatch::Http::Headers Class'
+          end
```
Fryguy: Minor style, but it's preferable to use guard clauses for validation...that is, this format is preferable...

```ruby
def self.validate_headers(val)
  raise ArgumentError, 'Not an ActionDispatch::Http::Headers Class' unless val.is_a?(ActionDispatch::Http::Headers)
  true
end
```

Fryguy: Oh, also, should this probably be a private class method?

syncrou: Ha! - I thought our rubocop was against `unless` - I wanted to make it (had it) that way in the first place.

This makes me happy.  😄

Fryguy:
@mkanoor Made a great point about thread local cleanup and the with_user style, but I don't see it here.  Was that expected in a follow up?  Also, @mkanoor, is the with_user thing only used in tests in ManageIQ, or do they use that in the controllers as well?

syncrou:
@Fryguy - Yes, all the user handling is in a followup.  I've built most of it already from this branch.

syncrou:
@Fryguy - Adjusted the validate method, now makes use of `unless` and is a `private_class_method`

mkanoor:
@Fryguy The with_user is used in controllers, in the queue dispatch etc because we want to clear out the thread local. So in this use case would the controller just call the with_request method which takes a block.

Issue #20, opened 2018-12-18
Add Identity struct/class


Fryguy:
This would be a module that extracts Identity from the headers and creates an object that is accessible at  the Thread level - See related #12

Fryguy:
Closed via #31

PR #31, opened 2019-01-30
Add the ManageIQ::API::Common::User class for managing a user hash in the header

syncrou:
1. Decode and validate a user pulled from `ManageIQ::API::Common::Request.current` hash

Using the below user header representation, the `ManageIQ::API::Common::Request.user` method returns the current user object to the block with all keys ( except 'type' ) nested under `identity` represented as static methods on the object.
```
user = {
  "identity"=> {
    "account_number" => "0369233",
    "type" => "User",
    "user" => {
      "username" => "jdoe",
      "email" => "jdoe@acme.com",
      "first_name" => "John",
      "last_name" => "Doe",
      "is_active" => true,
      "is_org_admin" => false,
      "is_internal" => false,
      "locale" => "en_US"
    },
    "internal" => {
      "org_id" => "3340851",
      "auth_type" => "basic-auth",
      "auth_time" => 6300
     }
  }
}
```

Testing via a console session:

```
2.4.1 :001 > rh_header = {"x-rh-identity"=>"eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjAzNjkyMzMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJqZG9lIiwiZW1haWwiOiJqZG9lQGFjbWUuY29tIiwiZmlyc3RfbmFtZSI6IkpvaG4iLCJsYXN0X25hbWUiOiJEb2UiLCJpc19hY3RpdmUiOnRydWUsImlzX29yZ19hZG1pbiI6ZmFsc2UsImlzX2ludGVybmFsIjpmYWxzZSwibG9jYWxlIjoiZW5fVVMifSwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMzM0MDg1MSIsImF1dGhfdHlwZSI6ImJhc2ljLWF1dGgiLCJhdXRoX3RpbWUiOjYzMDB9fX0="}
 => {"x-rh-identity"=>"eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjAzNjkyMzMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJqZG9lIiwiZW1haWwiOiJqZG9lQGFjbWUuY29tIiwiZmlyc3RfbmFtZSI6IkpvaG4iLCJsYXN0X25hbWUiOiJEb2UiLCJpc19hY3RpdmUiOnRydWUsImlzX29yZ19hZG1pbiI6ZmFsc2UsImlzX2ludGVybmFsIjpmYWxzZSwibG9jYWxlIjoiZW5fVVMifSwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMzM0MDg1MSIsImF1dGhfdHlwZSI6ImJhc2ljLWF1dGgiLCJhdXRoX3RpbWUiOjYzMDB9fX0="}

encoded_headers = ActionDispatch::Request.new(rh_header)

ManageIQ::API::Common::Request.current = encoded_headers2.4.1 :002 >
2.4.1 :003 > encoded_headers = ActionDispatch::Request.new(rh_header)
 => #<ActionDispatch::Request:0x00007fae632ec598 @env={"x-rh-identity"=>"eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjAzNjkyMzMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJqZG9lIiwiZW1haWwiOiJ
qZG9lQGFjbWUuY29tIiwiZmlyc3RfbmFtZSI6IkpvaG4iLCJsYXN0X25hbWUiOiJEb2UiLCJpc19hY3RpdmUiOnRydWUsImlzX29yZ19hZG1pbiI6ZmFsc2UsImlzX2ludGVybmFsIjpmYWxzZSwibG9jYWxlIjoiZW5fVVMifSwiaW50ZXJuYWwiOnsib3J
nX2lkIjoiMzM0MDg1MSIsImF1dGhfdHlwZSI6ImJhc2ljLWF1dGgiLCJhdXRoX3RpbWUiOjYzMDB9fX0="}, @filtered_parameters=nil, @filtered_env=nil, @filtered_path=nil, @protocol=nil, @port=nil, @method=nil, @re
quest_method=nil, @remote_ip=nil, @original_fullpath=nil, @fullpath=nil, @ip=nil>

ManageIQ::API::Common::Request.current = encoded_headers2.4.1 :004 >
2.4.1 :005 > ManageIQ::API::Common::Request.current = encoded_headers
 => #<ActionDispatch::Request:0x00007fae632ec598 @env={"x-rh-identity"=>"eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjAzNjkyMzMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJqZG9lIiwiZW1haWwiOiJqZG9lQGFjbWUuY29tIiwiZmlyc3RfbmFtZSI6IkpvaG4iLCJsYXN0X25hbWUiOiJEb2UiLCJpc19hY3RpdmUiOnRydWUsImlzX29yZ19hZG1pbiI6ZmFsc2UsImlzX2ludGVybmFsIjpmYWxzZSwibG9jYWxlIjoiZW5fVVMifSwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMzM0MDg1MSIsImF1dGhfdHlwZSI6ImJhc2ljLWF1dGgiLCJhdXRoX3RpbWUiOjYzMDB9fX0="}, @filtered_parameters=nil, @filtered_env=nil, @filtered_path=nil, @protocol="http://", @port=80, @method=nil, @request_method=nil, @remote_ip=nil, @original_fullpath="", @fullpath="", @ip=nil, @headers=#<ActionDispatch::Http::Headers:0x00007fae623cdc60 @req=#<ActionDispatch::Request:0x00007fae632ec598 ...>>>

2.4.1 :006 > request = ManageIQ::API::Common::Request.current
 => #<ManageIQ::API::Common::Request:0x00007fae623cd300 @headers=#<ActionDispatch::Http::Headers:0x00007fae623cdc60 @req=#<ActionDispatch::Request:0x00007fae632ec598 @env={"x-rh-identity"=>"ey
JpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjAzNjkyMzMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJqZG9lIiwiZW1haWwiOiJqZG9lQGFjbWUuY29tIiwiZmlyc3RfbmFtZSI6IkpvaG4iLCJsYXN0X25hbWUiOiJEb2UiLCJpc1
9hY3RpdmUiOnRydWUsImlzX29yZ19hZG1pbiI6ZmFsc2UsImlzX2ludGVybmFsIjpmYWxzZSwibG9jYWxlIjoiZW5fVVMifSwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMzM0MDg1MSIsImF1dGhfdHlwZSI6ImJhc2ljLWF1dGgiLCJhdXRoX3RpbWUiOjYzMD
B9fX0="}, @filtered_parameters=nil, @filtered_env=nil, @filtered_path=nil, @protocol="http://", @port=80, @method=nil, @request_method=nil, @remote_ip=nil, @original_fullpath="", @fullpath="",
 @ip=nil, @headers=#<ActionDispatch::Http::Headers:0x00007fae623cdc60 ...>>>, @original_url="://::80">

request.user.username2.4.1 :007 >
2.4.1 :008 > request.user.username
 => "jdoe"
```


syncrou:
@bzwei @hsong-rh @mkanoor - This is wip, as I still need to add tests and need to implement the `with_user` || `with_request` method, but wouldn't mind your feed back early on to see if this is what you can use for Approval etc.

lib/manageiq/api/common/users.rb
```diff
@@ -0,0 +1,70 @@
+module ManageIQ
+  module API
+    module Common
+      module Users
+        IDENTITY_KEY = 'x-rh-identity'.freeze
+        VALID_KEYS = %w(account_number user username email is_org_admin)
```
mkanoor: @syncrou Do we need this list if we decode it we get back a hash and when the user asks for a key if it exists in the hash we give the value back else we raise a key not found error.

lib/manageiq/api/common/users.rb
```diff
@@ -0,0 +1,70 @@
+module ManageIQ
+  module API
+    module Common
+      module Users
```
mkanoor: @syncrou 
Should this just be a class called Users and we return an instance of the user after we have parsed the headers.
The user instance would store the decoded hash and allow the user to access the key value pairs

lib/manageiq/api/common/users.rb
```diff
@@ -0,0 +1,70 @@
...
+        IDENTITY_KEY = 'x-rh-identity'.freeze
+        VALID_KEYS = %w(account_number user username email is_org_admin)
+
+        def self.current
```
mkanoor: @syncrou This should return back a user object after it has parsed out the identity key from the header.
Preferably stored in a thread local variable

lib/manageiq/api/common/users.rb
syncrou: Ahh yes, we were discussing an `OpenStruct` class to return the keys as attributes.

lib/manageiq/api/common/users.rb
```diff
@@ -0,0 +1,70 @@
...
+          Thread.current[:attr_current_user] = validate
+        end
+
+        def self.with_user(other_headers)
```
mkanoor: @syncrou I think this should take a user object and a code block

```
def with_user(user = User.current_user, block)
end
```

lib/manageiq/api/common/users.rb
```diff
@@ -0,0 +1,70 @@
...
+
+        def self.decode(options = nil, key = IDENTITY_KEY)
+          headers = options ? options : ManageIQ::API::Common::Headers.current
+          raise StandardError, "Requires a valid header object" unless headers
+
+          val = headers.instance_variable_get(:@req).stringify_keys
+          JSON.parse(Base64.decode64(val[key]))
+        end
+
+        def self.encode(val)
```
mkanoor: @syncrou I don't think this function is needed, whatever 3 scale generates will be based on the IPP we can't have the clients just adding arbitrary keys into the hash

lib/manageiq/api/common/users.rb
```diff
@@ -0,0 +1,70 @@
...
+        def self.current
+          Thread.current[:attr_current_user] = validate
+        end
+
+        def self.with_user(other_headers)
+          # Not Implemented
+          raise "Not Implemented"
+        end
+
+        def self.decode(options = nil, key = IDENTITY_KEY)
```
mkanoor: @syncrou I think this has to be a private method

lib/manageiq/api/common/users.rb
syncrou: @mkanoor - This ended up here partly because of this comment: https://github.com/ManageIQ/topological_inventory-api/pull/104#discussion_r251944320

syncrou:
Addresses https://github.com/ManageIQ/manageiq-api-common/issues/20

syncrou:
@mkanoor @bzwei @Fryguy please review

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,60 @@
...
+  module API
+    module Common
+      class User
+        attr_reader :identity
+
+        IDENTITY_KEY = 'x-rh-identity'.freeze
+        VALID_USER_HEADER_KEYS = %w(account_number user username email is_org_admin)
+
+        def self.current
+          Thread.current[:attr_current_user] = new
```
mkanoor: @syncrou 
should this be Thread.current[:attr_current_user] ||= new

How would this get reset?
Do we need a with_user method?

lib/manageiq/api/common/headers.rb
```diff
@@ -11,6 +11,15 @@ def self.current
           Thread.current[:attr_current_headers]
         end
 
+        def self.encode(val)
```
mkanoor: @syncrou Who would use this, we are not building any new headers we are just forwarding headers we receive from 3scale to other services

lib/manageiq/api/common/user.rb
syncrou: Regarding the `with_user` method - https://github.com/ManageIQ/manageiq-api-common/pull/31#issuecomment-459059377 ( Yes )

spec/lib/manageiq/api/common/headers_spec.rb
```diff
@@ -10,4 +10,14 @@
       expect { ManageIQ::API::Common::Headers.current = header_bad }.to raise_exception(ArgumentError)
     end
   end
+
+  context "Base64 encoding" do
+    it 'returns a Base64 representation of a hash' do
+      headers = { 'x-rh-auth-identity' => { 'identity' => { 'is_org_admin' => false } } }
```
mkanoor: @syncrou This needs to be x-rh-identity based on https://docs.google.com/document/d/1PAzJqcUXlxg7t5cX1lsPsQtBPT_95bZbyB9Iiv_ekGM/edit#heading=h.9rvb6r5r6f4n

lib/manageiq/api/common/headers.rb
syncrou: @Fryguy asked for this on a Topology PR, at least to have a helper method to handle it. I put it in  Headers so as to not lose that conversation.  Do you see a need for `self.encode` @Fryguy?

spec/lib/manageiq/api/common/user_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+#      "locale": "en_US"
+#    },
+#    "internal" : {
+#      "org_id": "3340851",
+#      "auth_type": "basic-auth",
+#      "auth_time": 6300
+#     }
+#  }
+describe ManageIQ::API::Common::User do
+  let(:encoded) { {"x-rh-identity"=>"eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjAzNjkyMzMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJqZG9lIiwiZW1haWwiOiJqZG9lQGFjbWUuY29tIiwiZmlyc3RfbmFtZSI6IkpvaG4iLCJsYXN0X25hbWUiOiJEb2UiLCJpc19hY3RpdmUiOnRydWUsImlzX29yZ19hZG1pbiI6ZmFsc2UsImlzX2ludGVybmFsIjpmYWxzZSwibG9jYWxlIjoiZW5fVVMifSwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMzM0MDg1MSIsImF1dGhfdHlwZSI6ImJhc2ljLWF1dGgiLCJhdXRoX3RpbWUiOjYzMDB9fX0="} }
```
mkanoor: @syncrou Should we use a Factory for this and create normal_user and admin user traits?

lib/manageiq/api/common/user.rb
syncrou: Yes to the other comment too, I'll get that change out.

syncrou:
@mkanoor @bzwei applied feedback, please review.
@Fryguy please review.

lib/manageiq/api/common/headers.rb
Fryguy: I had asked for a common "encode this specific identity structure and put it in a header" method.  I don't see a need for exposing the generic encoding of any arbitrary hash.

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,63 @@
...
+        ensure
+          Thread.current[:attr_current_user] = nil
+        end
+
+        def initialize(user = nil)
+          @current = user || ManageIQ::API::Common::Headers.current
+          user_hash = validate
+          @identity = user_hash['identity']
+          build_methods
+          self
```
Fryguy: This line is not needed (i.e. initialize doesn't return anything so having this here is a no-op)

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,63 @@
...
+          if user.keys.include?('identity') && user['identity'].keys.include?('user')
+            return user if (identity_list & VALID_USER_HEADER_KEYS) == VALID_USER_HEADER_KEYS
+          end
+          raise StandardError, "Not a valid user header hash"
+        end
+
+        def build_methods
+          user = flattened_header
+          user.keys.each do |key|
+            self.class.send(:define_method, :"#{key}") do
```
Fryguy: Why are we dynamically creating anything?

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,63 @@
...
+        VALID_USER_HEADER_KEYS = %w(account_number user username email is_org_admin)
+
+        def self.with_logged_in_user(user = nil)
+          yield Thread.current[:attr_current_user] = new(user)
+        ensure
+          Thread.current[:attr_current_user] = nil
+        end
+
+        def initialize(user = nil)
+          @current = user || ManageIQ::API::Common::Headers.current
```
Fryguy: This reads funny... I read this as "user or headers", which aren't related at all.  Is this being initialized with header data or a user object?

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,63 @@
+module ManageIQ
+  module API
+    module Common
+      class User
+        IDENTITY_KEY = 'x-rh-identity'.freeze
+        VALID_USER_HEADER_KEYS = %w(account_number user username email is_org_admin)
+
+        def self.with_logged_in_user(user = nil)
+          yield Thread.current[:attr_current_user] = new(user)
```
Fryguy: Not sure what the point of the thread current is, if you're not actually creating a direct accessor for it.

Is with_logged_in_user the only was to use this?  Can you give an example how this would work in a controller, because I can't envision how it will work

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,63 @@
...
+          user.keys.each do |key|
+            self.class.send(:define_method, :"#{key}") do
+              user[key]
+            end
+          end
+        end
+
+        def flattened_header
+          user_hash = @identity['user']
+          user_hash.merge!(@identity['internal'])['account_number'] = @identity['account_number']
```
Fryguy: It feels bad to modify data that's not owned by this module (i.e. you are effectively editing the header here)

syncrou: Yes. I ran into a bug due to the same thing in the tests.  I'll get that cleaned up.

lib/manageiq/api/common/user.rb
mkanoor: @syncrou @Fryguy 

We are trying to model this after the User.with_user in ManageIQ so that setting of the Thread Local Storage (TLS) and unsetting of the TLS happens in a single function and a block gets passed in.

https://github.com/ManageIQ/manageiq/blob/0648257adeab79e217a4d7783b961108f113e797/app/models/user.rb#L263

So in the controllers for every action I am thinking we would have

def some_action
   User.with_logged_in_user do
      call_service_object
      render :json ...
  end
end

lib/manageiq/api/common/user.rb
bzwei: Can we just explicitly create needed methods? Return `nil` for optional attributes. We can have more meaningful method names such as `#tenancy`, or we can hide some attributes for example `internal`.

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,63 @@
...
+            val = @current.stringify_keys
+          end
+          JSON.parse(Base64.decode64(val[key]))
+        end
+
+        def validate
+          check_user(decode)
+        end
+
+        def check_user(user)
```
bzwei: The validation code is a little hard to read because the VALID_USER_HEADER_KEYS is a mix of top level and second level keys. 
If we explicitly construct attribute methods we can move validation to each individual method. Raise an error if a mandatory attribute is missing; return `nil` if an optional attribute is missing.

syncrou:
@Fryguy @mkanoor @bzwei All feedback has been applied.
Please review.

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,79 @@
...
+          yield alternate_user
+        ensure
+          alternate_user = nil
+        end
+
+        def initialize(user_header)
+          @user = user_header
+        end
+
+        def username
```
mkanoor: @syncrou How would this work if the 3 scale team or the RHN login adds a new field. Would we have to update this code or would we have direct access to the raw hash to handle on-going enhancements from outside of us.

In the IPP there is a nested hash is they key names get duplicated in the nested hash what would these functions return?

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,79 @@
...
+        def self.current
+          Thread.current[:attr_current_user]
+        end
+
+        def self.current=(user_header)
+          Thread.current[:attr_current_user] = new(user_header)
+        end
+
+        def self.with_logged_in_user(user_header)
+          alternate_user = new(user_header)
```
mkanoor: @syncrou Shouldn't this use the self.current= and set the value? as opposed to using alternate user.

This call will typically get used by the Background Queue process e.g. When Provisioning gets an approval from Approval Service Message Bus it would fetch the saved headers from the DB and call this function to set the User.current to mimic as if it came from the Controller.

lib/manageiq/api/common/user.rb
bzwei: @mkanoor I advocate we explicitly expose selected attributes rather than blindly pass through the raw hash. The consumer of this class should rely on the class contract rather than the backend (3scale or RHN) document. 

@syncrou shouldn't `tenant` be exposed too? I saw your existing work directly examined the header to fetch tenant. Is it better to get tenant through the user object?

lib/manageiq/api/common/user.rb
syncrou: @mkanoor - I don't have any issue using `self.current=`, though it seemed there was no use for the `Thread.current` (TLS) when used inside a block so I decided to use a local var instead.  Would it be better to set `Thread.current` and ensure it is nilled out like my original code?

lib/manageiq/api/common/user.rb
syncrou: @bzwei - Yeah, I'll get that added.

lib/manageiq/api/common/user.rb
mkanoor: @syncrou We are going to be writing methods that will use
the User.current_user for example when sending out Webhook payload or calling Approval Service with the name of the user that created the request. So we don't have a choice we we have to set the TLS so that the underlying methods be consistent when they are called from the Rails Controller or a background job engine like Sidekiq where there is no logged in user but we have access to the previously stored headers in the Order.

syncrou: @mkanoor - I'll get that change in.

syncrou:
@mkanoor, @bzwei Feedback has been applied.

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,87 @@
+module ManageIQ
+  module API
+    module Common
+      class User
+        IDENTITY_KEY = 'x-rh-identity'.freeze
+
+        def self.current
+          Thread.current[:attr_current_user]
```
mkanoor: @syncrou Should we raise an error here if the TLS is not set?

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,87 @@
...
+        rescue NoMethodError => e
+          raise ArgumentError, "Not a valid user hash: #{e.message}"
+        end
+
+        def validate_and_return(key)
+          find_key(decode, key)
+        end
+
+        def find_key(hash, key)
+          hash['identity'].has_key?(key.to_s) ? hash['identity'][key.to_s] : hash['identity']['user'][key.to_s]
```
bzwei: To address @mkanoor's earlier concern about the case where the first level `['identity']` and the second level `['identity']['user']` contain the same key, we'd better to separate them clearly.
Consider rename this method name to `find_user_key`, and let `tenant` method not calling into this method but directly checking `hash['identity']['account_number']`.

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,87 @@
...
+
+        def self.current=(user_header)
+          Thread.current[:attr_current_user] = new(user_header)
+        end
+
+        def self.with_logged_in_user(user_header)
+          self.current = user_header
+          yield self.current
+        ensure
+          Thread.current[:attr_current_user] = nil
```
mkanoor: @syncrou 
Should this be self.current = nil instead of directly setting the TLS?

syncrou: I can call `self.current = nil` but I have to then check that method to see if nil is passed in.  Otherwise new is always called. Do we want `self.current` to have the ability to be set as `nil`?

lib/manageiq/api/common/user.rb
syncrou: I'll get that change in

lib/manageiq/api/common/user.rb
Fryguy: Note the method from more_core_extensions that would be helpful: `has_key_path?`

lib/manageiq/api/common/user.rb
Fryguy: It feels to me like the `current=` setter maybe should not take a raw header, because it's inconsistent with the getter which returns a wrapped object. If we make the getter return a wrapped object, and the setter take a wrapped object, then we either need another setter (like `raw_current=` and/or some kind of factory method `User.from_header`).  Once all that's in place, `current=nil` should be straightforward.

lib/manageiq/api/common/user.rb
Fryguy: This feels easier to read: `hash.fetch_path('identity', key.to_s) || hash.fetch_path('identity', 'user', key.to_s)`

What do you think?

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,87 @@
...
+
+        def tenant
+          validate_and_return(:account_number)
+        end
+
+        private
+
+        def decode
+          if @user.is_a?(ActionDispatch::Http::Headers)
+            val = @user.instance_variable_get(:@req).stringify_keys
```
Fryguy: The instance_variable_get feels weird to me.

Fryguy: I saw something in @mkanoor 's PR, and I think this code will actually break outside of tests as is.  When you use `ActionDispatch::Http::Headers.new`, it's expecting a request object as the parameter, not a Hash.  As such, pulling out `@req` in production is invalid (might work by happenstance, but not the right approach).

Instead of using `.new` in the tests, you can use `.from_hash`, and then you can just call `ActionDispatch::Http::Headers#to_h` after that to get the headers back out.  The .to_h only works when you initialize it in this more accurate way.

Fryguy:
Sorry...I wasn't clear in how I described the raw_current, so let me redescribe it...in addition, I noticed that we shouldn't be duplicating the Headers because they already exist in the headers object, and we should just use that directly (I apologize if the original code was this way...I can't remember).  In psuedocode, I was expecting something like...

```ruby
class User
  def initialize(headers)
    # ... extract from the hash
  end

  def self.current
    return Thread[:attr_current_user] if Thread[:attr_current_user]
    raise "Headers have not been set" if Headers.current.nil?
    self.current = new(Headers.current)
  end

  def self.current=(user)
    Thread[:attr_current_user] = user
  end

  def self.with_user(user)
    self.current = user
    yield
  ensure
    self.current = nil
  end
end
```

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,105 @@
...
+        def is_internal?
+          validate_and_return(:is_internal)
+        end
+
+        def locale
+          validate_and_return(__method__)
+        end
+
+        def tenant
+          result = decode.dig('identity','account_number')
```
Fryguy: Surprised rubocop didn't complain here, but needs a space after comma.

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,105 @@
...
+          raise StandardError "Tenant key doesn't exist" if result.nil?
```
Fryguy: No idea how this is working, since the comma is missing...I would have expected a SyntaxError.

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,105 @@
...
+
+        private
+
+        def decode
+          if @user.is_a?(ActionDispatch::Http::Headers)
+            val = @user.instance_variable_get(:@req).stringify_keys
+          else
+            val = @user.stringify_keys
+          end
+          JSON.parse(Base64.decode64(val[IDENTITY_KEY]))
```
Fryguy: I think we have to care about case-sensitivity here with the lookup from the IDENTITY_KEY because HttpHeaders are case sensitive.  I think the `ActionDispatch::Http::Headers` already handles the case-insensitive nature already, so flipping this around might work better...

```
def decode
  user = ActionDispatch::Http::Headers.from_hash(user) if user.is_a?(Hash)
  JSON.parse(Base64.decode64(user[IDENTITY_KEY]))

...
```

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,105 @@
...
+        rescue NoMethodError => e
```
Fryguy: What causes a NoMethodError?  I can't understand how this would occur.

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,105 @@
...
+          raise ArgumentError, "Not a valid user hash: #{e.message}"
+        end
+
+        def validate_and_return(key)
+          find_user_key(decode, key)
+        end
+
+        def find_user_key(hash, key)
+          result = hash.dig('identity', 'user', key.to_s)
+          raise StandardError "#{key} doesn't exist" if result.nil?
```
Fryguy: Also missing a comma...or maybe this is valid syntax?

lib/manageiq/api/common/user.rb
Fryguy: Dug into this, and it's a total coincidence this isn't a SyntaxError.  This code, if you try to execute it, gives...

```ruby
irb(main):001:0> def aaa
irb(main):002:1> raise StandardError "xxx"
irb(main):003:1> end
irb(main):004:0> aaa
NoMethodError: undefined method `StandardError' for main:Object
    from (irb):2:in `aaa'
    from (irb):4
    from /home/bdunne/.rubies/ruby-2.4.5/bin/irb:11:in `<main>'
```

I think specs will catch this.  Does this also explain the NoMethodError question I had elsewhere?

lib/manageiq/api/common/user.rb
syncrou: It (rubocop) isn't running at all for me on this repo.  I realized that earlier this week.

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,105 @@
...
+        def self.from_header(raw_header)
+          self.raw_current = raw_header
+          raw_header.nil? ? Thread.current[:attr_current_user] = nil : self.current = new(raw_current)
+        end
+
+        def self.with_logged_in_user(user_header)
+          self.from_header(user_header)
+          yield self.current
+        ensure
+          Thread.current[:attr_current_user] = nil
```
Fryguy: Saw this pattern in @mkanoor 's PR as well, and he did the following which I think is preferable...

```ruby
        def self.with_logged_in_user(user_header)
          saved_user = current
          self.from_header(user_header)
          yield current
        ensure
          self.current = saved_user
        end
```

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,105 @@
...
+        end
+
+        def self.current=(user)
+          raise StandardError, "Must be a ManageIQ::API::Common::User" unless user.is_a?(ManageIQ::API::Common::User)
+          Thread.current[:attr_current_user] = user
+        end
+
+        def self.from_header(raw_header)
+          self.raw_current = raw_header
+          raw_header.nil? ? Thread.current[:attr_current_user] = nil : self.current = new(raw_current)
```
Fryguy: from_header shouldn't set the current, or at least to a casual reader they would never expect that side effect.  I would expect a method named from_header to essentially just be a factory method that returns a User object form a header.

That being said, I don't think this is needed as all based on my other comment.

lib/manageiq/api/common/user.rb
syncrou: it "raises an exception for keys that do not exist" do
      bad_user.each do |key|
        expect{ user.send(key)}.to raise_exception(StandardError)
      end
    end
I wonder how this test was passing.  Odd.

syncrou:
@mkanoor @bzwei if you could take another pass at this that would be great.

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,69 @@
...
+          raise StandardError, "Tenant key doesn't exist" if result.nil?
+          result
+        end
+
+        private
+
+        def decode
+          hashed = ManageIQ::API::Common::Request.current.to_h
+          user_hash = hashed[:headers][IDENTITY_KEY]
+          JSON.parse(Base64.decode64(user_hash))
```
bzwei: Do we have to actual decode the user_hash every time since decode is so often called in this class? We may cache the decoded result.
There is a possibility that current user and current request are out of sync by caching the result, but the real use case for current user is almost within the context of current request.

Fryguy: +1.  This should memoize the decoding because it's using the "current" header (and for performance).

spec/lib/manageiq/api/common/request_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  context "user" do
+    it "returns a lazy loaded user instance" do
+      described_class.with_request(request_good) do |instance|
+        expect(instance.user).to be_a(ManageIQ::API::Common::User)
+      end
+    end
+  end
+
+  context "set current headers" do
```
Fryguy: I think you also need a test where the with_request is called with a Hash to a) test that code path and b) show that we can reconstruct this object in non-controller environments.

spec/lib/manageiq/api/common/request_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    it 'raises an exception if the class is incorrect' do
+      expect do
+        described_class.with_request(request_bad) {}
+      end.to raise_exception(ArgumentError)
+    end
+  end
+
+  context "get forwardable headers" do
```
Fryguy: Can you make this match the method name under test?  i.e. `describe ".current_forwardable"`

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,65 @@
...
+        end
+
+        def validate_and_return(key)
+          find_user_key(decode, key)
+        end
+
+        def find_user_key(hash, key)
+          result = hash.dig('identity', 'user', key.to_s)
+          raise StandardError, "#{key} doesn't exist" if result.nil?
```
Fryguy: Should there be a more specific error class?

lib/manageiq/api/common/user.rb
```diff
@@ -0,0 +1,65 @@
...
+        end
+
+        def locale
+          validate_and_return(__method__)
+        end
+
+        def tenant
+          result = decode.dig('identity', 'account_number')
+          raise StandardError, "Tenant key doesn't exist" if result.nil?
+          result
```
Fryguy: If we are going to have a `find_user_key` method, then we probably should have a `find_tenant_key` as well for parity.

spec/spec_helper.rb
```diff
@@ -20,7 +20,7 @@
   config.mock_with :rspec do |mocks|
     mocks.verify_partial_doubles = true
   end
-
+  config.include UserHeaderSpecHelper
```
Fryguy: can you move this to the top of the configure block so that it stands out a bit?  Here in the middle it's sort of buried amongst the boilerplate configuration and you can't see it.

Fryguy:
Lookin pretty good...I agree with @bzwei 's comment, and the rest is more specs for more cases, and other minor things.

lib/manageiq/api/common/user.rb
syncrou: Memoization will be added in a follow up PR as per this issue and conversations with @Fryguy: https://github.com/ManageIQ/manageiq-api-common/issues/36

syncrou:
@Fryguy @bzwei all feedback applied.

PR #33, opened 2019-02-13
Pagination fixes

bdunne:
- Return relative links for pagination
- Only return link references that have values in order to be swagger compliant


bdunne:
Fixes: https://github.com/ManageIQ/topological_inventory-api/issues/125

lib/manageiq/api/common/paginated_response.rb
```diff
@@ -31,7 +31,7 @@ def links_hash
             "last"  => link_to_last,
             "prev"  => link_to_prev,
             "next"  => link_to_next,
-          }
+          }.compact
```
Fryguy: whoa...I didn't know that would remove nil values... we should deprecate `delete_nils` from more_core_extensions.

PR #34, opened 2019-02-26
Support for forwardable headers

mkanoor:
When service portal or approval service needs to communicate
with other services, it needs to forward certain headers to the
other services.


mkanoor:
@bdunne @Fryguy 
Please review

lib/manageiq/api/common/headers.rb
```diff
@@ -1,7 +1,9 @@
+require 'byebug'
```
bzwei: To be removed

lib/manageiq/api/common/headers.rb
```diff
@@ -11,6 +12,11 @@ def self.current
           Thread.current[:attr_current_headers]
         end
 
+        def self.forwardable_headers
+          raise "Current Headers has not been set" unless Thread.current[:attr_current_headers]
+          Thread.current[:attr_current_headers].to_h.slice(*FORWARDABLE_HEADER_KEYS)
```
Fryguy: Minor, but this should use the accessor instead of going directly to the Thread local...

```ruby
        def self.forwardable_headers
          raise "Current Headers has not been set" unless current
          current.to_h.slice(*FORWARDABLE_HEADER_KEYS)
        end
```

lib/manageiq/api/common/headers.rb
```diff
@@ -11,6 +12,11 @@ def self.current
           Thread.current[:attr_current_headers]
         end
 
+        def self.forwardable_headers
+          raise "Current Headers has not been set" unless Thread.current[:attr_current_headers]
```
Fryguy: `"current header have not been set"`

Also wondering if this should raise a specific error other than RuntimeError, or perhaps not raise an exception at all, and instead just return `{}`

spec/lib/manageiq/api/common/headers_spec.rb
```diff
@@ -10,4 +10,12 @@
       expect { ManageIQ::API::Common::Headers.current = header_bad }.to raise_exception(ArgumentError)
     end
   end
+
+  context "get forwardable headers" do
+    it "x-rh-identity" do
+      ManageIQ::API::Common::Headers.current = header_good
+      hash = ManageIQ::API::Common::Headers.forwardable_headers
+      expect(hash['x-rh-identity']).to eq('abc')
```
Fryguy: Should this test include `X-Request-ID` as well?

spec/lib/manageiq/api/common/headers_spec.rb
```diff
@@ -10,4 +10,12 @@
       expect { ManageIQ::API::Common::Headers.current = header_bad }.to raise_exception(ArgumentError)
     end
   end
+
+  context "get forwardable headers" do
+    it "x-rh-identity" do
+      ManageIQ::API::Common::Headers.current = header_good
+      hash = ManageIQ::API::Common::Headers.forwardable_headers
```
Fryguy: Minor, but you can use `described_class` instead of typing out the entire class name.  Useful for if it ever gets renamed.

spec/lib/manageiq/api/common/headers_spec.rb
Fryguy: You probably should also have a test that `:blah` is *not* included in the hash.

Fryguy:
@mkanoor A few minor comments, but overall looks good!

lib/manageiq/api/common/headers.rb
```diff
@@ -11,6 +12,11 @@ def self.current
           Thread.current[:attr_current_headers]
         end
 
+        def self.forwardable_headers
```
Fryguy: `Headers.forwardable_headers` seems redundant...perhaps `Headers.forwardable`?

Fryguy: Same with const name

lib/manageiq/api/common/headers.rb
```diff
@@ -2,6 +2,7 @@ module ManageIQ
   module API
     module Common
       module Headers
+        FORWARDABLE_HEADER_KEYS = %w(X-Request-ID x-rh-identity)
```
Fryguy: Do we care about case sensitivity here?  header keys are case-insensitive.

lib/manageiq/api/common/headers.rb
bdunne: We'll need this in order to return a useful error document

mkanoor:
@bdunne @Fryguy 
Please review

lib/exceptions.rb
```diff
@@ -0,0 +1,7 @@
+module ManageIQ
+  module API
+    module Common
+      class HeadersNotSet < StandardError; end
```
bdunne: Should this inherit from `ArgumentError`?  The headers are sort of like arguments to the request

lib/manageiq/api/common/headers.rb
```diff
@@ -2,17 +2,30 @@ module ManageIQ
...
 
         def self.current
           Thread.current[:attr_current_headers]
         end
 
+        def self.with_headers(headers)
+          self.current = headers
+          yield current
+        ensure
+          self.current = nil
```
bdunne: Was it `nil` or some other value before?

lib/manageiq/api/common/headers.rb
```diff
@@ -2,17 +2,30 @@ module ManageIQ
   module API
     module Common
       module Headers
+        FORWARDABLE_HEADER_KEYS = %w(X-Request-ID x-rh-identity).freeze
         def self.current=(val)
-          validate_headers(val)
+          validate_headers(val) if val
           Thread.current[:attr_current_headers] = val
```
bdunne: Should this be removed `if val == nil`?

mkanoor: @bdunne  we are trying to unset a thread variable so we need to set it to a nil it to a nil and valdiate_headers rejects a nil because it is expecting a header object. So either the ensure block can directly set the thread variable to a nil or go thru current=

lib/exceptions.rb
```diff
@@ -0,0 +1,7 @@
+module ManageIQ
+  module API
+    module Common
+      class HeadersNotSet < ArgumentError; end
```
Fryguy: This exceptions file name should also be namespaced, otherwise you run the risk of this colliding with other gems that have an exceptions at the top level... `require 'manageiq/api/common/exceptions'`

lib/manageiq-api-common.rb
```diff
@@ -1 +1,2 @@
 require 'manageiq/api/common'
+require 'exceptions'
```
Fryguy: this file is just used as a redirect to the main require file.  So, this line really belongs in `lib/manageiq/api/common.rb`

lib/manageiq/api/common/headers.rb
```diff
@@ -2,17 +2,31 @@ module ManageIQ
...
+        def self.with_headers(headers)
+          saved_headers = current
+          self.current = headers
+          yield current
+        ensure
+          self.current = saved_headers
+        end
+
+        def self.current_forwardable
+          raise ManageIQ::API::Common::HeadersNotSet, "Current header have not been set" unless Thread.current[:attr_current_headers]
```
Fryguy: `unless current`... not sure why we wouldn't use the accessor here.

spec/lib/manageiq/api/common/headers_spec.rb
```diff
@@ -1,13 +1,36 @@
...
-      expect { ManageIQ::API::Common::Headers.current = header_bad }.to raise_exception(ArgumentError)
+      expect do
+        described_class.with_headers(header_bad) {}
+      end.to raise_exception(ArgumentError)
+    end
+  end
+
+  context "get forwardable headers" do
+    it "x-rh-identity" do
+      described_class.with_headers(header_good) do |_x|
```
Fryguy: Minor, but the block var is not needed at all...that is you can just do `do` without the `|_x|`

Fryguy:
@mkanoor A couple more minor comments on the new changes...everything else looks great so far!

lib/manageiq-api-common.rb
bdunne: If the file is properly namespaced, the require shouldn't be necessary since autoload should handle it.

PR #35, opened 2019-03-01
Add limit and offset to PaginatedResponse meta.

martinpovolny:
It's convenient to have all the paging metainformation available in every paginated responses.

Adding the `limit` and `offset` fields.

meta 2019-03-01_13-01

Ping @Hyperkid123

### DIFF
--- a/lib/manageiq/api/common/paginated_response.rb
+++ b/lib/manageiq/api/common/paginated_response.rb
@@ -22,7 +22,7 @@ module ManageIQ
         private
 
         def metadata_hash
-          @metadata_hash ||= {"count" => count}
+          @metadata_hash ||= {"count" => count, "limit" => limit, "offset" => offset}
         end
 
         def links_hash
--- a/spec/lib/manageiq/api/common/paginated_response_spec.rb
+++ b/spec/lib/manageiq/api/common/paginated_response_spec.rb
@@ -23,6 +23,28 @@ describe ManageIQ::API::Common::PaginatedResponse do
     end
   end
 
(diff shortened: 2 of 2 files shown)


martinpovolny:
@bdunne : guess I need to add some tests. And I see where.

Is there anything else that needs to be done?

Fryguy:
@mkanoor also had a similar issue as he did not now how to use clients that don't accept URLs directly (most clients don't...they just expect a method call, with optional paramaters directly), and the only thing available in the "next" link was a full link.  This PR should solve this issue.

Hyperkid123:
+1 for this PR i would love to have these attributes in pagination responses so i can actually use the client methods.

lib/manageiq/api/common/paginated_response.rb
```diff
@@ -22,7 +22,7 @@ def response
         private
 
         def metadata_hash
-          @metadata_hash ||= {"count" => count}
+          @metadata_hash ||= {"count" => count, "limit" => @limit, "offset" => @offset}
```
bdunne: The `@`'s shouldn't be needed since there are `attr_reader`s for both `limit` and `offset`

bdunne:
Would you mind adding some tests for this in https://github.com/ManageIQ/manageiq-api-common/blob/master/spec/lib/manageiq/api/common/paginated_response_spec.rb ?  Otherwise, looks great :+1:

martinpovolny:
Thx!

Removed `@`s and added 2 simple tests.

PR #37, opened 2019-03-04
Fixed a bug in the key names

mkanoor:
~When creating HTTP header keys if you are using `from hash` 
the key names should be normalized, all dashed converted to
underscores. Or we have to use the 'add' function which prefixes
everything with HTTP_~

~https://github.com/rails/rails/blob/94b5cd3a20edadd6f6b8cf0bdf1a4d4919df86cb/actionpack/lib/action_dispatch/http/headers.rb#L122~

EDIT @Fryguy - See https://github.com/ManageIQ/manageiq-api-common/pull/37#issuecomment-469774810 for changes to this PR after discussion with @mkanoor


Fryguy:
Is this all-caps-and-http thing only for specs? or will Request#to_h have to return it this way so we can round_trip via forwarding?

mkanoor:
@Fryguy Its only for the spec since we are using the from_hash method.

Fryguy:
Here's what we are thinking:

```ruby
class Request
  def initialize(headers: , original_url: )
    @headers = from_hash(headers) if headers.is_a?(Hash)
    # ...
  end

  def to_h
    {:headers => forwardable, :original_url => original_url}
  end

  private

  def forwardable
    # if dev, only extract jwt token, else extract x-rh-identity and x-request-id
    # extraction is done via the loop logic: FORWARDABLE_HEADER_KEYS.each_with_object({}) do |key, hash|
    # since jwt is in a cookie, this may mean cracking open the cookie jar, pulling out that one cookie and mapping it to a different header x-rh-jwt?
  end

  def from_hash(hash)
    ActionDispatch::Http::Headers.from_hash({}).tap do |headers|
      hash.each { |k, v| headers.add(k, v) }
    end
  end
end

**Comment by bzwei** [comment]
@syncrou seems you need to change the value in https://github.com/ManageIQ/manageiq-api-common/blob/master/lib/manageiq/api/common/user.rb#L5 to `HTTP_X_RH_IDENTITY` too.

File: lib/manageiq/api/common/request.rb [hunk 1]
```diff
@@ -47,7 +47,21 @@ def user
         end
 
         def to_h
-          {:headers => headers.to_h, :original_url => original_url}
+          {:headers => forwardable, :original_url => original_url}
+        end
+
+        def forwardable
+          FORWARDABLE_HEADER_KEYS.each_with_object({}) do |key, hash|
```
bdunne: Any reason not to use the `.slice` from above?

lib/manageiq/api/common/request.rb
```diff
@@ -47,7 +47,21 @@ def user
         end
 
         def to_h
-          {:headers => headers.to_h, :original_url => original_url}
+          {:headers => forwardable, :original_url => original_url}
+        end
+
+        def forwardable
```
bdunne: Should this be private?

lib/manageiq/api/common/request.rb
Fryguy: Yes.  It turns out that .to_h of an ActionDispatch::Http::Headers doesn't do what you'd expect.  We need to use the accessors instead.

Fryguy: More accurately `ActionDispatch::Http::Headers.from_hash` doesn't do what you'd expect.  That is, from_hash expects a Hash of what look like ENV vars.  Methods are expecting that structure in order to work in a case-insensitive manner.  This is the best examples I could come up with:

```ruby
irb(main):003:0> h = ActionDispatch::Http::Headers.from_hash("x-foo" => "x")
=> #<ActionDispatch::Http::Headers:0x00007f92405a0760 @req=#<ActionDispatch::Request:0x00007f92405a0788 @env={"x-foo"=>"x"} ...>
irb(main):004:0> h["x-foo"]
=> nil
irb(main):005:0> h["HTTP_X_FOO"]
=> nil
irb(main):006:0> h = ActionDispatch::Http::Headers.from_hash("HTTP_X_FOO" => "x")
=> #<ActionDispatch::Http::Headers:0x00007f92405ae798 @req=#<ActionDispatch::Request:0x00007f92405ae7c0 @env={"HTTP_X_FOO"=>"x"} ...>
irb(main):007:0> h["x-foo"]
=> "x"
irb(main):008:0> h["HTTP_X_FOO"]
=> "x"
```

Fryguy:
> @syncrou seems you need to change the value in https://github.com/ManageIQ/manageiq-api-common/blob/master/lib/manageiq/api/common/user.rb#L5 to HTTP_X_RH_IDENTITY too.

That should not be necessary, as the purpose of the to_h method is to avoid that inside-knowledge.

lib/manageiq/api/common/request.rb
Fryguy: It can't be because we call it up in the self.current_forwardable method

PR #38, opened 2019-03-05
Removing unused/confusing config directory

Fryguy:
The config directory, as is, makes it look like that's where you have to config the repo for tests, but then it doesn't work.  Removing it entirely as it's not used.

Also changed the name in the sample file so that when working across multiple applications they are not all named "dummy_*"


spec/dummy/config/database.sample.yml
```diff
@@ -10,14 +10,14 @@ base: &base
 
 development:
```
syncrou: @Fryguy when using `bin/console` it uses `config/database.yml` so this change makes running tests easier, but confuses the same for `bin/console`  Is there any fix that we could use for both? Or possibly a better error message letting the user know what isn't configured correctly

Fryguy: I did `bundle exec rails console` and it worked ... can you try that?

syncrou: That works.  I didn't know `rails c` was available.  I've been using `bin/console`. Probably means `bin/console` can be removed.

PR #39, opened 2019-03-05
Refactor user accessors

Fryguy:
@syncrou What do you think?

This trims up the repetition, makes find_user_key work on decode directly the same way find_tenant does, and changes the `?` accessors to not have the leading `is_`

### DIFF
--- a/lib/manageiq/api/common/user.rb
+++ b/lib/manageiq/api/common/user.rb
@@ -4,36 +4,19 @@ module ManageIQ
       class User
         IDENTITY_KEY = 'x-rh-identity'.freeze
 
-        def username
-          validate_and_return(__method__)
-        end
-
-        def email
-          validate_and_return(__method__)
-        end
-
-        def first_name
-          validate_and_return(__method__)
-        end
-
-        def last_name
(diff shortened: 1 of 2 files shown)


lib/manageiq/api/common/user.rb
```diff
@@ -4,36 +4,19 @@ module Common
...
-        end
-
-        def is_internal?
-          validate_and_return(:is_internal)
-        end
-
-        def locale
-          validate_and_return(__method__)
+        %w[
+          username
```
syncrou: Would it make sense to move this list into a constant?

Fryguy: I don't believe so, because it won't be used any other place than here, and putting it in a constant a) exposes it outside of this class and b) doesn't allow it to be GC'd

Fryguy:
Not doing the rubocop because we are changing that at the core .rubocop.yml level.

syncrou:
@bzwei - Is this going to break anything currently in use for Approval?

bzwei:
@syncrou No. The change has no impact on approval service.

PR #43, opened 2019-03-07
Rename HeadersNotSet to RequestNotSet

Fryguy:
Additionally add a helper method Request.current! that raises when then
Request.current is not set.

@syncrou Please review.  If there is any calling code actually catching HeadersNotSet, it would have to be changed for this PR.  If needed I can add a temporary alias of the constant.
cc @bdunne

### DIFF
--- a/lib/manageiq/api/common/request.rb
+++ b/lib/manageiq/api/common/request.rb
@@ -1,11 +1,21 @@
 module ManageIQ
   module API
     module Common
+      class RequestNotSet < ArgumentError
+        def initialize
+          super("Current request has not been set")
+        end
+      end
+
       class Request
         FORWARDABLE_HEADER_KEYS = %w(X-Request-ID x-rh-identity).freeze
 
         def self.current
           Thread.current[:current_request]
+        end
+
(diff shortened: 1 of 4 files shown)


lib/manageiq/api/common/request.rb
```diff
@@ -1,13 +1,24 @@
...
+
       class Request
         FORWARDABLE_HEADER_KEYS = %w(X-Request-ID x-rh-identity).freeze
 
         def self.current
           Thread.current[:current_request]
         end
 
+        def self.current!
+          raise RequestNotSet unless current
```
bdunne: What do you think of this instead?
```ruby
def self.current!
  current || raise(RequestNotSet)
end
```

Fryguy: I'd love to but that's invalid syntax.  This is that specific case where you have to use `or` - `current or raise RequestNotSet`.  Would you want that?

Fryguy: Turns out the parens are required syntax.

PR #44, opened 2019-03-07
Add accessors for headers

### LINKED ISSUE #36: Add Memoization for User accessor methods

Fryguy:
This PR adds accessors for specific header keys in the Request, and by doing so separates the "header" knowledge from the User class.  That is, the user just gets the identity hash from the request and doesn't have to know it's a header.

@syncrou Please review

Built on #43 so just review the second commit.
Closes #36

### DIFF
--- a/lib/manageiq/api/common/request.rb
+++ b/lib/manageiq/api/common/request.rb
@@ -8,7 +8,9 @@ module ManageIQ
       end
 
       class Request
-        FORWARDABLE_HEADER_KEYS = %w(X-Request-ID x-rh-identity).freeze
+        REQUEST_ID_KEY = "x-request-id".freeze
+        IDENTITY_KEY   = 'x-rh-identity'.freeze
+        FORWARDABLE_HEADER_KEYS = [REQUEST_ID_KEY, IDENTITY_KEY].freeze
 
         def self.current
           Thread.current[:current_request]
@@ -46,13 +48,21 @@ module ManageIQ
 
         attr_reader :headers, :original_url
 
-        def initialize(headers:, original_url:, **kwargs)
+        def initialize(headers:, original_url:, **_kwargs)
           headers = from_hash(headers) if headers.kind_of?(Hash)
           @headers, @original_url = headers, original_url
         end
 
+        def request_id
+          headers.fetch(REQUEST_ID_KEY)
+        end
+
+        def identity
+          @identity ||= JSON.parse(Base64.decode64(headers.fetch(IDENTITY_KEY)))
+        end
+
         def user
-          @user ||= User.new
+          @user ||= User.new(identity)
         end
 
         def to_h
@@ -61,7 +71,7 @@ module ManageIQ
 
         def forwardable
           FORWARDABLE_HEADER_KEYS.each_with_object({}) do |key, hash|
-            hash[key] = @headers[key] if @headers.key?(key)
+            hash[key] = headers[key] if headers.key?(key)
           end
         end
 
--- a/lib/manageiq/api/common/user.rb
+++ b/lib/manageiq/api/common/user.rb
@@ -1,8 +1,12 @@
 module ManageIQ
   module API
     module Common
+      class IdentityError < StandardError; end
+
       class User
-        IDENTITY_KEY = 'x-rh-identity'.freeze
+        def initialize(identity)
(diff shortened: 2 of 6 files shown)


lib/manageiq/api/common/request.rb
```diff
@@ -31,17 +44,24 @@ def self.with_request(request)
...
         def self.current_forwardable
-          raise ManageIQ::API::Common::HeadersNotSet, "Current headers have not been set" unless current
-          current.forwardable
+          current!.forwardable
         end
 
         attr_reader :headers, :original_url
 
-        def initialize(headers:, original_url:, **kwargs)
+        def initialize(headers:, original_url:, **_kwargs)
```
bzwei: should we remove the unused `**_kwargs`?

lib/manageiq/api/common/request.rb
```diff
@@ -31,17 +44,24 @@ def self.with_request(request)
...
           headers = from_hash(headers) if headers.kind_of?(Hash)
           @headers, @original_url = headers, original_url
         end
 
+        def request_id
+          headers[REQUEST_ID_KEY]
+        end
+
+        def identity
+          JSON.parse(Base64.decode64(headers[IDENTITY_KEY]))
```
bzwei: why not cached the identity here rather than in each user instance?
should raise an error if headers does not have key `IDENTITY_KEY`.

lib/manageiq/api/common/request.rb
```diff
@@ -31,17 +44,24 @@ def self.with_request(request)
...
+        end
+
         def user
           @user ||= User.new
```
bzwei: Can we return `nil` if header does not have `IDENTITY_KEY`? The other option is to raise an error.

lib/manageiq/api/common/request.rb
Fryguy: This is there to capture and "extra" parameters that might be sent along, otherwise the caller blows up.  We can probably remove it if we are sure all callers are doing the right thing.  I'd leave that to a separate PR, though.

lib/manageiq/api/common/request.rb
Fryguy: I like the idea of caching it here.  I'll do that.

If the identity key is missing, should this raise or just return nil.  I was thinking just return nil, but I'm not sure.

lib/manageiq/api/common/request.rb
Fryguy: I agree this code should do that as well (or perhaps User.new should do that and it would bubble up to here).

lib/manageiq/api/common/request.rb
syncrou: From an RBAC perspective if the user doesn't exist nothing can happen.  So maybe we should `raise` instead of return `nil`.

lib/manageiq/api/common/user.rb
```diff
@@ -1,9 +1,9 @@
 module ManageIQ
   module API
     module Common
-      class User
-        IDENTITY_KEY = 'x-rh-identity'.freeze
+      class IdentityError < StandardError; end
```
syncrou: Same question about errors

lib/manageiq/api/common/request.rb
```diff
@@ -1,13 +1,26 @@
 module ManageIQ
   module API
     module Common
+      class RequestNotSet < ArgumentError
+        def initialize
```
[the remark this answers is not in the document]
Fryguy: only that it is directly related to the request object.  I really wanted to move it *inside* the Request class (i.e. `ManageIQ::API::Common::Request::NotSet`), but this was more of a first step.

lib/manageiq/api/common/user.rb
Fryguy: Yup see https://github.com/ManageIQ/manageiq-api-common/pull/44#discussion_r263913837 ☝️

Fryguy:
Updated with the suggestions.  Also, based on the rationale that the User class can be isolated form the "headers", I've changed the signature of User.new to take an identity payload (not headers, just the guts of the identity payload).  This way we can also test it in isolation from the headers themselves.  I did not yet change the specs for user though to not use headers...I can do that in a follow up PR.

PR #45, opened 2019-03-08
Add more specs and cover missing calls

Fryguy:
Refactor specs to avoid the with_request repetition.  Also include more
headers in the example requests.

@bdunne @syncrou Please review.


Fryguy:
@syncrou @bdunne I'd like this merged first before my other 2 PRs, and then I will rebase those on top of this and include tests for the things I am changing along with the those PRs.

PR #48, opened 2019-03-28
Using the wrong header key for x-rh-insights-request-id

mkanoor:
We were incorrectly using x-request-id

The x-request-id comes from rails via this config setting

https://github.com/ManageIQ/catalog-api/blob/daa356d660efc78bcc5d5375a5917fbb463a3f0d/config/environments/production.rb#L48

### DIFF
--- a/lib/manageiq/api/common/request.rb
+++ b/lib/manageiq/api/common/request.rb
@@ -8,9 +8,9 @@ module ManageIQ
       end
 
       class Request
-        REQUEST_ID_KEY = "x-request-id".freeze
+        X_RH_INSIGHTS_REQUEST_ID_KEY = "x-rh-insights-request-id".freeze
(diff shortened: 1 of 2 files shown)


lib/manageiq/api/common/request.rb
```diff
@@ -8,9 +8,9 @@ def initialize
       end
 
       class Request
-        REQUEST_ID_KEY = "x-request-id".freeze
+        X_RH_INSIGHTS_REQUEST_ID_KEY = "x-rh-insights-request-id".freeze
```
Fryguy: can we just keep the constant named REQUST_ID_KEY...it's more readable, and there's no need to make a constant exactly equal to it's value.

Fryguy: also the method name is still request_id, so it keeps that parity.

mkanoor:
@Fryguy please review

Issue #10, opened 2018-12-18
Add Versioning DSL for routing


Fryguy:
I'm thinking the usage in the app routes file would be roughly something like

```ruby
Rails.application.routes.draw do
  ManageIQ::API::Versioning.versioned_routes(") do
    resources :whatevers
  end
end
```

where the module is defined something like

```ruby
module ManageIQ
  module API
    module Versioning
      def self.versioned_routes(version)
        #...
        scope :as => :api, :module => "api", :path => prefix do
          match "/#{version.split(".").first}/*path", :via => [:delete, :get, :options, :patch, :post, :put], :to => redirect("#{prefix}/#{version}/%{path}")
          namespace version.sub(".", "x").to_sym, :path => version do
            yield
          end
        end
        #...
      end
    end
  end
end
```

PR #50, opened 2019-04-03
Add a class to encapsulate the non-trivial redirect for versioning

carbonin:
~~I have no idea how to test this without adding in the whole versioning scaffolding to the dummy app.~~

~~Any ideas @Fryguy ?~~

Added tests.


carbonin:
This is part of the fix for #10, but doesn't handle the whole route scope.

lib/manageiq/api/common/action_dispatch-routing-option_redirect.rb
```diff
@@ -0,0 +1,23 @@
+module ManageIQ
```
Fryguy: Minor, but I think this file should be renamed and/or the class renamed to match.  I can merge w/o this though.

carbonin: Renamed the file.

PR #53, opened 2019-04-10
Raise IdentityError if x-rh-identity doesn't exist on User initialize

lindgrenj6:
Instead of letting the `KeyError` blow through here and subsequently rescuing for something as generic as KeyError, raise `ManageIQ::API::Common::IdentityError` instead since there isn't an identity to begin with due to the lack of an `x-rh-identity` header. 

On the catalog side we've been rescuing from a `KeyError` for a long time, but this makes much more sense and is more descriptive. It also allows us to change the rails behavior on `ActionController::ParameterMissing` since that is a KeyError itself it was hard to handle it any differently application-wide.

### DIFF
--- a/lib/manageiq/api/common/request.rb
+++ b/lib/manageiq/api/common/request.rb
@@ -63,6 +63,8 @@ module ManageIQ
 
         def identity
           @identity ||= JSON.parse(Base64.decode64(headers.fetch(IDENTITY_KEY)))
+        rescue KeyError
+          raise IdentityError, "x-rh-identity not found"
         end
 
         def user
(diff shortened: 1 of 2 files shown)


lindgrenj6:
@miq-bot add_reviewer gmcculloug 
@miq-bot add_reviewer syncrou
@miq-bot add_reviewer Fryguy 
@miq-bot add_reviewer mkanoor

lindgrenj6:
This is related to https://github.com/ManageIQ/catalog-api/pull/222, makes more sense to have a custom exception here.

PR #55, opened 2019-04-17
Allow .current= argument to be a ManageIQ::API::Common::Request type

bzwei:
Fix an issue for
https://github.com/ManageIQ/manageiq-api-common/pull/55/files#diff-35dc214cbf96bc055529320b09ba5ee3R46 where we need to restore from saved current which is a type of `ManageIQ::API::Common::Request` if not `nil`, but the setter does not accept this type.

### DIFF
--- a/lib/manageiq/api/common/request.rb
+++ b/lib/manageiq/api/common/request.rb
@@ -31,10 +31,10 @@ module ManageIQ
               new(:headers => request.headers, :original_url => request.original_url)
             when Hash
               new(request)
-            when nil
+            when Request, nil
               request
             else
-              raise ArgumentError, 'Not an ActionDispatch::Http::Headers Class or Hash, or nil'
+              raise ArgumentError, 'Not a ManageIQ::API::Common::Request or ActionDispatch::Request Class, Hash, or nil'
             end
         end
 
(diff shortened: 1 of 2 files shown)


mkanoor:
@bzwei 
Did this happen because you have nested calls meaning
```
ManageIQ::API::Common::Request.with_request(request1) do
   ManageIQ::API::Common::Request.with_request(request2) do
     ManageIQ::API::Common::Request.with_request(request3) do
     end
  end
end
```

bzwei:
@mkanoor Yes. Each controller already starts with a scope of request, then inside any business logic where we need to switch a request context we will hit this issue.

mkanoor:
@bzwei I understand this is a bug. But why would your business logic switch the context it should just work with the one that is set and carry it thru to the end. Are you fabricating contexts or are you using contexts that was set once. What we have seen is that the context needs to be just set once in the beginning for all paths coming from the controller and once in the background job for an event coming in we read the context stored in the database and set it for the current background job thread.

bzwei:
@mkanoor We need to switch context because we need to act as admin (or system) to call other service, for example, to call RBAC to access groups.

PR #56, opened 2019-04-18
Print a warning when no spec dummy database.yml exists

syncrou:
When running tests locally for the common gem, the default warning doesn't reference the path
where the `database.yml` file is missing.
This error message loads from `spec_helper.rb` and stops the specs
unless the file exists. It also references where the `database.yml` file should be placed.


syncrou:
- cc @lindgrenj6

PR #57, opened 2019-04-23
Add documentation for Filter parameter format

lindgrenj6:
Add some filter examples so we can refer to here from catalog/approval/topology/sources etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,25 @@ Or install it yourself as:
 
 ## Usage
 
-TODO: Write usage instructions here
+#### ManageIQ::Api::Common::Filter
+
+After implementing filtering in your application, this is the way to filter via parameters on index functions:
+
+| Query Parameter | Ruby Client Parameter | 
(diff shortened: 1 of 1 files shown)


gmcculloug:
@abellotti Do you like or do you like?  😃

abellotti:
for this I'd prefer to see a 2 column table showing the "Query Parameter" for the first and "Ruby Client Parameter" for the second. it would be easier to see.  Thanks!!

lindgrenj6:
@abellotti updated!

README.md
```diff
@@ -29,6 +29,26 @@ Or install it yourself as:
 
 TODO: Write usage instructions here
 
+#### ManageIQ::Api::Common::Filter
```
syncrou: We can probably remove the TODO on line 30 since you have instructions here for one part of the gem.

lindgrenj6: done!

PR #59, opened 2019-04-26
Add entitlements

syncrou:
Adds `Entitlement` class allowing services to query current enabled / disabled entitlements on a service.

This PR does _not_ raise the `EntitlementError` even though it is setup inline with the class. Until Entitlements are checked in all services, we are allowing all requests through if the `entitlements` key is not found.
```
  "entitlements": {
    "hybrid_cloud": {
      "is_entitled": true
    },
    "insights": {
      "is_entitled": true
    },
    "openshift": {
      "is_entitled": true
    },
    "smart_management": {
      "is_entitled": true
    }
  }
}
```
Examples:
```
2.4.1 :001 >  ent_hash =  {"entitlements" => {
2.4.1 :002 >           "hybrid_cloud" => {
2.4.1 :003 >               "is_entitled" => true
2.4.1 :004?>           },
2.4.1 :005 >           "insights" => {
2.4.1 :006 >               "is_entitled" => true
2.4.1 :007?>           },
2.4.1 :008 >           "openshift" => {
2.4.1 :009 >               "is_entitled" => true
2.4.1 :010?>           },
2.4.1 :011 >           "smart_management" => {
2.4.1 :012 >               "is_entitled" => true
2.4.1 :013?>           }
2.4.1 :014?>       }
2.4.1 :015?>   }
 => {"entitlements"=>{"hybrid_cloud"=>{"is_entitled"=>true}, "insights"=>{"is_entitled"=>true}, "openshift"=>{"is_entitled"=>true}, "smart_management"=>{"is_entitled"=>true}}}

ents = ManageIQ::API::Common::Entitlement.new(ent_hash)2.4.1 :016 >
2.4.1 :017 > ents = ManageIQ::API::Common::Entitlement.new(ent_hash)
 => #<ManageIQ::API::Common::Entitlement:0x00007ff98c123be8 @identity={"entitlements"=>{"hybrid_cloud"=>{"is_entitled"=>true}, "insights"=>{"is_entitled"=>true}, "openshift"=>{"is_entitled"=>true}, "smart_management"=>{"is_entitled"=>true}}}>
2.4.1 :018 > ents.openshift?
 => true
2.4.1 :019 > ents.hybrid_cloud?
 => true
2.4.1 :020 > ents.smart_management?
 => true
2.4.1 :021 > ents.insights
Traceback (most recent call last):
        1: from (irb):21
NoMethodError (undefined method `insights' for #<ManageIQ::API::Common::Entitlement:0x00007ff98c123be8>)
Did you mean?  insights?
2.4.1 :022 > ents.insights?
=> true
```

### DIFF
--- a/lib/manageiq/api/common/entitlement.rb
+++ b/lib/manageiq/api/common/entitlement.rb
@@ -25,7 +25,7 @@ module ManageIQ
 
         def find_entitlement_key(key)
           result = identity.dig('entitlements', key.to_s)
-          #TODO Always force entitlements key
+          # TODO: Always force entitlements key
           return true unless result
           result['is_entitled']
         end
--- a/lib/manageiq/api/common/entitlement.rb
+++ b/lib/manageiq/api/common/entitlement.rb
@@ -1,0 +1,35 @@
+module ManageIQ
+  module API
+    module Common
+      class EntitlementError < StandardError; end
+
+      class Entitlement
+        def initialize(identity)
+          @identity = identity
+        end
+
+        %w[
+          hybrid_cloud
+          insights
+          openshift
+          smart_management
+        ].each do |m|
+          define_method(m) do
+            find_entitlement_key(m)
+          end
+        end
+
+        private
+
+        attr_reader :identity
+
+        def find_entitlement_key(key)
+          result = identity.dig('entitlements', key.to_s)
+          return true unless result
+          #raise EntitlementError, "#{key} doesn't exist" if result.nil?
+          result['is_entitled']
+        end
+      end
+    end
+  end
+end
--- a/lib/manageiq/api/common/entitlement.rb
+++ b/lib/manageiq/api/common/entitlement.rb
@@ -14,7 +14,7 @@ module ManageIQ
           openshift
           smart_management
         ].each do |m|
-          define_method(m) do
+          define_method("#{m}?") do
             find_entitlement_key(m)
           end
         end
@@ -25,8 +25,8 @@ module ManageIQ
 
         def find_entitlement_key(key)
           result = identity.dig('entitlements', key.to_s)
+          #TODO Always force entitlements key
           return true unless result
-          #raise EntitlementError, "#{key} doesn't exist" if result.nil?
           result['is_entitled']
         end
       end
--- a/spec/lib/manageiq/api/common/entitlements_spec.rb
+++ b/spec/lib/manageiq/api/common/entitlements_spec.rb
@@ -9,8 +9,8 @@ describe ManageIQ::API::Common::Entitlement do
   end
 
   context "entitlement getter methods" do
-    let(:entitlement_keys) { %w(insights openshift hybrid_cloud smart_management) }
-    let(:bad_entitlement)  { %w(fred barney type) }
+    let(:entitlement_keys) { %w[insights openshift hybrid_cloud smart_management] }
+    let(:bad_entitlement)  { %w[fred barney type] }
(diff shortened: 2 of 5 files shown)


lib/manageiq/api/common/entitlement.rb
```diff
@@ -0,0 +1,35 @@
+module ManageIQ
+  module API
+    module Common
+      class EntitlementError < StandardError; end
+
+      class Entitlement
+        def initialize(identity)
```
mkanoor: @syncrou Can this default to picking up the identity from the Request object and raise an error if its nil.
Is this expected to be called from the base controllers for every micro service?

syncrou: Eventually it will, but until other micro services add the check for entitlements we're allowing everything through unless the entitlements hash exists. If it exists we raise exceptions if it is invalid.

https://projects.engineering.redhat.com/browse/SSP-327

gmcculloug:
@Fryguy @gtanzillo Could you help review or assign others?  Thanks.

lib/manageiq/api/common/entitlement.rb
```diff
@@ -0,0 +1,35 @@
...
+          @identity = identity
+        end
+
+        %w[
+          hybrid_cloud
+          insights
+          openshift
+          smart_management
+        ].each do |m|
```
abellotti: why is this list hardcoded ? I'd expect this to be dynamic based on the identity.entitlements content, minimizing churn in this code as the entitlement service adds things in the future. The dynamic methods can be created during initialization based on the identity.

syncrou: Hi @abellotti, it's hardcoded per the request by Jason to do the same on https://github.com/ManageIQ/manageiq-api-common/blob/master/lib/manageiq/api/common/user.rb#L11-L24

My thought was the churn here will need to be managed anyway. So controlling it instead of new elements showing up unannounced makes sense.

Fryguy: I also don't think we should hardcode this, as it's likely to change frequently, unlike the identity keys, which are likely static.

That being said, I'm ok with merging for now, but I think this particular bit should be dynamic.

spec/lib/manageiq/api/common/entitlements_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  context "entitlement getter methods" do
+    let(:entitlement_keys) { %w[insights openshift hybrid_cloud smart_management] }
+    let(:bad_entitlement)  { %w[fred barney type] }
+    let(:entitlement)      { ManageIQ::API::Common::Request.current.entitlement }
+    let(:other_user)       { default_user_hash }
+
+    it "returns values for entitlement methods" do
+      entitlement_keys.each do |key|
+        expect(entitlement.respond_to?("#{key}?")).to be_truthy
```
Fryguy: spec-wise, this can be hard to find issues when something goes wrong because rspec will stop on the failed key, but it won't show you which key is failing.  It's usually better to loop the other way around...

```ruby
entitlement_keys.each do |key|
  it "returns values for #{key} entitlement" do
    expect(entitlement.respond_to?("#{key}?")).to be_truthy
  end
end
```

Fryguy:
All of my comments are minor and could be in a follow-up, so merging.

lib/manageiq/api/common/entitlement.rb
bzwei: I agree with @syncrou's thought. Explicitly managing the elements make the code easy to read and use. Do we really expect major products are added to the list frequently?

PR #63, opened 2019-05-06
If the x-rh-insights-request-id is missing return nil

mkanoor:
Currently it validates the request to contain the header key
x-rh-insights-request-id if its missing it raises the KeyError
Suppressing it to return a nil if the key doesn't exist


mkanoor:
@Fryguy @agrare 
Please review

PR #69, opened 2019-05-28
Adding common support for GraphQL.

abellotti:
Adding common support for GraphQL.

This includes:
  - GraphQL Schema generator that is based on OpenAPI
  - The necessary schema templates used
  - The additional schema types used
  - Helper methods for the GraphQL controller

Moving this GraphQL Common logic from https://github.com/ManageIQ/topological_inventory-api/pull/194


abellotti:
With this PR.  GraphQL can be quickly enabled in OpenAPI based micro-services.
- Testing in progress with:
  - Topological Inventory API
  - Sources API

abellotti:
/cc @Fryguy as we chatted yesterday, before I forget them, these are the main enhancements:

- Expose the versioned GraphQL paths /api/<service>/v#.#/graphql
- Move from rake triggered GraphQL schema generation to JIT GraphQL Schema generation based on accessed path above.
  - Remove the [graphql] option from the service's openapi:generate tasks
  - GraphQL Schema generation can be done on the fly from the graphql controller via a call like ::Api::Docs[version].init_graphql
- Ditch the Relay Node interface for the more simpler GraphQL core signature for paging.
  - Implement the offset and limit parameters in the primary collections
  - Example:

Relay Specification:

```
{
  sources(after: "MQ", first: 5) {
    edges {
      node {
        id
        name
        endpoints {
          edges {
            node {
              id
              schema
              host
              port
              authentications {
                edges {
                  node {
                    id
                    username
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

GraphQL Proper:

```
{
  sources(offset: 6, limit: 5) {
    id
    name
    endpoints {
      id
      schema
      host
      port
      authentications {
        id
        username
      }
    }
  }
}
```

Which is tons better, we'd just need to add a queryable pagination element that's more appropriate for GraphQL  (vs. REST's first, last, prev, next), i.e. page_size, has_next_page, number_of_pages.

abellotti:
Updated the GraphQL Prototype in manageiq-api-common and related PRs for sources and topological-inventory to take care of the first 2 items discussed with @Fryguy above. 

so ....

- GraphQL endpoints exposed is now the versioned path
  - /api/topological-inventory/v1.0/graphql
  - /api/sources/v1.0/graphql

  The openapi JSON has been updated to reflect that the graphql endpoints exist and this can be seen in the swagger editor too.

- This now implements a JIT GraphQL Schema generator.
  - No need to run openapi generate task with the graphql option.
  - The lib/api/graphql.rb schema file is no longer needed.
  - GraphQL Schema is dynamically generated for the exposed graphql endpoint's version
     so /api/sources/v1.0/graphql will create a v1.0 GraphQL schema based on the openapi v1.0 spec.

/cc @gtanzillo

gmcculloug:
cc @Hyperkid123

PR #75, opened 2019-06-06
Fix NoMethodError (undefined method `stats' for nil:NilClass)

bdunne:
When the Puma server is not initialized (doing anything other than rails s) the @get_stats variable on the Puma module is not defined, so calling stats on it raises NoMethodError.

### DIFF
--- a/lib/manageiq/api/common/middleware/web_server_metrics.rb
+++ b/lib/manageiq/api/common/middleware/web_server_metrics.rb
@@ -40,7 +40,7 @@ module ManageIQ
             @request_counter.increment(counter_labels)
             @request_histogram.observe(duration, duration_labels)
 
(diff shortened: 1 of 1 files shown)


lib/manageiq/api/common/middleware/web_server_metrics.rb
```diff
@@ -49,6 +49,12 @@ def call(env)
           def strip_ids_from_path(path)
             path.gsub(%r{/\d+(/|$)}, '/:id\\1')
           end
+
+          def puma_stats
+            JSON.parse(Puma.stats)
```
carbonin: I think I'd prefer `Puma ? JSON.parse(Puma.stats) : {}` over the rescue.

carbonin: Nevermind talked with @bdunne that's not the issue here. :tired_face:

PR #78, opened 2019-06-11
Drop relay based spec for base GraphQL query structure with filter/offset/limit

abellotti:
Drop support for the relay based specification and implements the base GraphQL query structure.
- Adding support for the following on collections:
  - offset
  - limit
  - filter

- And the following on associations:
  - offset
  - limit

Example with topological inventory:

```
{
  container_groups(filter: { name: { starts_with: "sources"}}) {
    id
    name
    containers (offset: 90, limit: 30) {
      id
      name
      archived_at
    }
    tags (offset: 20, limit: 20) {
      id
      description
    }
  }
}
```

Example with sources:

```
{
  sources(offset: 10, limit: 10, filter: { name: "OpenShift"}) {
    id
    name
    endpoints {
      id
      scheme
      host
      port
      path
      authentications {
        id
        username
        status
      }
    }
  }
}
```

### DIFF
--- a/lib/manageiq/api/common/graphql.rb
+++ b/lib/manageiq/api/common/graphql.rb
@@ -3,6 +3,8 @@ require "graphql/batch"
 require "graphql/preload"
 
 require "manageiq/api/common/graphql/generator"
+require "manageiq/api/common/graphql/association_loader"
+require "manageiq/api/common/graphql/associated_records"
 
 require "manageiq/api/common/graphql/types/big_int"
 require "manageiq/api/common/graphql/types/date_time"
--- a/lib/manageiq/api/common/graphql.rb
+++ b/lib/manageiq/api/common/graphql.rb
@@ -86,11 +86,15 @@ module ManageIQ
           }
         end
 
-        def self.paginated_search(scope, args)
-          offset = (args[:offset] || 0).to_i.clamp(0, Float::INFINITY)
-          limit  = (args[:limit]  || 100).to_i.clamp(1, 1000)
+        def self.ordered_search(scope, _args)
+          scope.order(:id)
+        end
+
+        def self.paged_search(scope, args)
           scope  = scope.where(:id => args[:id]) if args[:id]
-          scope.order(:id).offset(offset).limit(limit).all
+          offset = (args[:offset] ||   0).to_i.clamp(0, Float::INFINITY)
+          limit  = (args[:limit]  || 100).to_i.clamp(1, 1000)
+          scope.offset(offset).limit(limit)
         end
 
         # Following code is auto-generated via rails generate graphql:install
--- a/lib/manageiq/api/common/graphql.rb
+++ b/lib/manageiq/api/common/graphql.rb
@@ -2,10 +2,9 @@ require "graphql"
 require "graphql/batch"
 require "graphql/preload"
 
-require "manageiq/api/common/graphql/generator"
 require "manageiq/api/common/graphql/association_loader"
 require "manageiq/api/common/graphql/associated_records"
-
+require "manageiq/api/common/graphql/generator"
 require "manageiq/api/common/graphql/types/big_int"
 require "manageiq/api/common/graphql/types/date_time"
 require "manageiq/api/common/graphql/types/query_filter"
--- a/lib/manageiq/api/common/graphql.rb
+++ b/lib/manageiq/api/common/graphql.rb
@@ -86,6 +86,13 @@ module ManageIQ
           }
         end
 
+        def self.paginated_search(scope, args)
+          offset = (args[:offset] || 0).to_i.clamp(0, Float::INFINITY)
+          limit  = (args[:limit]  || 100).to_i.clamp(1, 1000)
+          scope  = scope.where(:id => args[:id]) if args[:id]
+          scope.order(:id).offset(offset).limit(limit).all
+        end
+
         # Following code is auto-generated via rails generate graphql:install
         #
         # Handle form data, JSON body, or a blank value
--- a/lib/manageiq/api/common/graphql.rb
+++ b/lib/manageiq/api/common/graphql.rb
@@ -88,15 +88,9 @@ module ManageIQ
           }
         end
 
-        def self.ordered_search(scope, _args)
-          scope.order(:id)
-        end
-
-        def self.paged_search(scope, args)
-          scope  = scope.where(:id => args[:id]) if args[:id]
-          offset = (args[:offset] ||   0).to_i.clamp(0, Float::INFINITY)
-          limit  = (args[:limit]  || 100).to_i.clamp(1, 1000)
-          scope.offset(offset).limit(limit)
+        def self.search_options(scope, args)
+          scope = scope.where(:id => args[:id]) if args[:id]
+          scope
         end
 
         # Following code is auto-generated via rails generate graphql:install
--- a/lib/manageiq/api/common/graphql.rb
+++ b/lib/manageiq/api/common/graphql.rb
@@ -89,8 +89,7 @@ module ManageIQ
         end
 
         def self.search_options(scope, args)
-          scope = scope.where(:id => args[:id]) if args[:id]
-          scope
+          args[:id] ? scope.where(:id => args[:id]) : scope
         end
 
         # Following code is auto-generated via rails generate graphql:install
--- a/lib/manageiq/api/common/graphql/associated_records.rb
+++ b/lib/manageiq/api/common/graphql/associated_records.rb
@@ -26,12 +26,12 @@ module ManageIQ
 
           def filter_collection_by_options(options)
             res = @collection
+            if options[:where].present?
+              options[:where].each_pair { |k, v| res = res.select { |rec| rec[k].to_s == v.to_s } }
+            end
             if options[:order].present?
               order_by = options[:order].first
               res = res.sort_by { |rec| rec[order_by] }
-            end
-            if options[:where].present?
-              options[:where].each_pair { |k, v| res = res.select { |rec| rec[k].to_s == v.to_s } }
             end
             res = res.drop(options[:offset]) if options[:offset]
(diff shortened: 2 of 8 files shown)


abellotti:
wanted this out early, so we can start leveraging the new signature (without updating queries later from edge/node to this).

todo in follow-up PRs,
- N+1
- Paging metadata

lib/manageiq/api/common/graphql.rb
```diff
@@ -80,6 +80,17 @@ def self.openapi_graphql_description
           }
         end
 
+        def self.ordered_search(scope, _args)
+          scope.order(:id)
+        end
+
+        def self.paged_search(scope, args)
```
bdunne: Any way we can share some of this code? https://github.com/ManageIQ/manageiq-api-common/blob/master/lib/manageiq/api/common/paginated_response.rb#L7

abellotti: No problem, was able to update GraphQL to actually leverage that class by exposing a helper method in there to expose the updated search scope. https://github.com/ManageIQ/manageiq-api-common/pull/78/commits/e1fcac7aabe7f7adb2c56238c2bd64e2e08e870f

so we now get the default common order, offset and limit set by that class.

lib/manageiq/api/common/graphql.rb
```diff
@@ -80,6 +82,11 @@ def self.openapi_graphql_description
           }
         end
 
+        def self.search_options(scope, args)
+          scope = scope.where(:id => args[:id]) if args[:id]
+          scope
```
Fryguy: No need for the temporal storage...

```ruby
args[:id] ? scope.where(:id => args[:id]) : scope
```

lib/manageiq/api/common/graphql/association_loader.rb
```diff
@@ -0,0 +1,35 @@
...
+          def perform(records)
+            records.each { |record| fulfill(record, read_association(record)) }
+          end
+
+          private
+
+          def read_association(record)
+            recs = ::ManageIQ::API::Common::GraphQL::AssociatedRecords.new(record.public_send(association_name).to_a)
+            recs = ::ManageIQ::API::Common::GraphQL.search_options(recs, args)
+            ::ManageIQ::API::Common::PaginatedResponse.new(
```
Fryguy: Can you remove the ManageIQ::API::Common from these lines?  I think the const resolution should work without that, and it woud be less verbose.

lib/manageiq/api/common/graphql/association_loader.rb
```diff
@@ -0,0 +1,35 @@
...
+
+          def perform(records)
+            records.each { |record| fulfill(record, read_association(record)) }
+          end
+
+          private
+
+          def read_association(record)
+            recs = ::ManageIQ::API::Common::GraphQL::AssociatedRecords.new(record.public_send(association_name).to_a)
+            recs = ::ManageIQ::API::Common::GraphQL.search_options(recs, args)
```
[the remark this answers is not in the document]
abellotti: I was really hoping for it, but In all 3 prototypes, graphql-preload/graphql-batch with association loader, anytime the where was done in the DB, the N+1 creeped back with GraphQL.  

If we didn't have to do any scoping (offset/limit/filter), the graphql-preload worked fine and no N+1, as soon as add a cope, bam, N+1.  I've tried specifying the scope with ActiveRecord Preloader, but that only made sense for common scopes for all resource associations, i.e. give me enabled records, or always id sorted, etc.  But that failed flat when we needed to offset/limit per association. (let alone filter).

Also played with the graphql-batch by itself, but that too was problematic as soon as I added DB scoping, anyways, see https://github.com/Shopify/graphql-batch/issues/77 and more particularly the comment in https://github.com/Shopify/graphql-batch/issues/77#issuecomment-461599811.  This lead me to the 
solution at hand with graphql-batch with the query-relation gem. i.e. Preload associations for all, (graphql-preload eliminated the N+1 just fine), then do the offset/limit/filter/order/... in ruby.   If there is another way I didn't try, by all means, let's try it.

lib/manageiq/api/common/paginated_response.rb
```diff
@@ -19,6 +19,10 @@ def response
           }
         end
 
+        def search
```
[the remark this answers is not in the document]
abellotti: 👍  I can just make the records accessor method public and use that.

lib/manageiq/api/common/graphql/associated_records.rb
```diff
@@ -0,0 +1,43 @@
...
+            when :first then res.first
+            when :last  then res.last
+            when :all   then res
+            end
+          end
+
+          def filter_collection_by_options(options)
+            res = @collection
+            if options[:order].present?
+              options[:order].each { |attr| res = res.sort_by { |rec| rec[attr] } }
```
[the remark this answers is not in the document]
abellotti: 👍 with query-relations options order came in as an array, I'll just do a first and use that.

lib/manageiq/api/common/graphql/association_loader.rb
```diff
@@ -0,0 +1,35 @@
...
+          end
+
+          def perform(records)
+            records.each { |record| fulfill(record, read_association(record)) }
+          end
+
+          private
+
+          def read_association(record)
+            recs = ::ManageIQ::API::Common::GraphQL::AssociatedRecords.new(record.public_send(association_name).to_a)
```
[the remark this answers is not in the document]
abellotti: Apparently, the .to_a is not needed for query-relation to work. will be removing that.

lib/manageiq/api/common/graphql/associated_records.rb
```diff
@@ -0,0 +1,44 @@
...
+
+module ManageIQ
+  module API
+    module Common
+      module GraphQL
+        class AssociatedRecords
+          include Enumerable
+          include QueryRelation::Queryable
+
+          def initialize(collection = [])
```
bdunne: Is there a case where this would be called without a collection?

manageiq-api-common.gemspec
```diff
@@ -29,9 +29,10 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "more_core_extensions"
 
   # For ManageIQ::API::Common::GraphQL
-  spec.add_runtime_dependency "graphql",         "~> 1.7"
-  spec.add_runtime_dependency "graphql-batch",   "~> 0.3.8"
-  spec.add_runtime_dependency "graphql-preload", "~> 1.0"
+  spec.add_runtime_dependency "graphql",         "~> 1.9"
+  spec.add_runtime_dependency "graphql-batch",   "~> 0.4"
+  spec.add_runtime_dependency "graphql-preload", "~> 2.0.1"
```
bdunne: Do they follow semver?  If so I would prefer to use `~> 2.0` with an additional `>= 2.0.1` if that's important.

lib/manageiq/api/common/graphql/associated_records.rb
bdunne: Actually... Is there a case where this wouldn't be an `ActiveRecord::Associations::CollectionProxy`?  With that in mind, should the `filter_collection_by_options` below use the database methods rather than doing everything in ruby?

abellotti: the N+1 problem come back when I do that. 😞

lib/manageiq/api/common/graphql/generator.rb
```diff
@@ -60,7 +60,7 @@ def self.collection_field_resolvers(schema_overlay, collection)
           end
 
           def self.init_schema(request, schema_overlay = {})
-            api_version       = ::ManageIQ::API::Common::GraphQL.version(request)
+            api_version       = GraphQL.version(request)
```
bdunne: Is this now referencing `::GraphQL` in the `graphql` gem?  If so, can https://github.com/ManageIQ/manageiq-api-common/pull/78/files#diff-bb4277a3c7abc4a5d84ffa1c301b5048R16 be removed?  If not, I think we should be explicit to avoid constant problems.

abellotti: was our own version, had updated for readability, but no problem, going back to fully qualified. 👍

abellotti:
ping @Fryguy anything else needed here ?

As discussed, I'll be adding the greatly needed GraphQL rspecs in a follow-up PR https://github.com/ManageIQ/manageiq-api-common/pull/85. Thanks.

Fryguy:
LGTM...if we merge will it break any existing applications?

Also, some conflicts and a couple of open questions from @bdunne

abellotti:
specs need updating after this is merged. so:

- https://github.com/ManageIQ/topological_inventory-api/pull/200
- https://github.com/ManageIQ/sources-api/pull/35

will need to be merged to fix those.

catalog-api will need something similar for the updated signature https://github.com/ManageIQ/catalog-api/blob/master/spec/requests/graphql_controller_spec.rb

/cc @syncrou

lib/manageiq/api/common/graphql/association_loader.rb
```diff
@@ -0,0 +1,35 @@
+module ManageIQ
+  module API
+    module Common
+      module GraphQL
+        class AssociationLoader < ::GraphQL::Batch::Loader
+          attr_reader :model, :association_name, :args
+
+          def initialize(model, association_name, args = {})
+            @model            = model
```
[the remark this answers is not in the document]
bdunne: This is just a minor nit, not important

lib/manageiq/api/common/graphql/associated_records.rb
bdunne: But, isn't it the same problem, just being handled in ruby instead of postgres?  It there are 10,000 subcollection items and the request has `limit: 5`, aren't we still going to instantiate 10,000 ruby objects, filter all 10,000, sort them then only return 5?

Issue #67, opened 2019-05-21
Validation of request parameters/body according to openapi v3 json


Ladas:
Lets design how to share validation in all of our APIs.

I have a POC, which is basically just testing that the validation gem I've picked works for ingress-api https://github.com/ManageIQ/topological_inventory-ingress_api/pull/47/files#diff-fdb60c134b290dc6cd1033a065180563R27

The gem I've picked and battle-tested is https://github.com/ota42y/openapi_parser/pull/33 (with 1 pending PR)

I've tested it on ingress openapi.json which is pretty complex https://github.com/ManageIQ/topological_inventory-ingress_api/blob/master/public/doc/openapi-3-v0.0.2.json

--------------------

Now how do we want to integrate this to all of our APIs? As a before_action for controller? Into rack middleware (existing project is https://github.com/interagent/committee but supports only 1 schema per server)?. More thoughts?

Ladas:
cc @Fryguy @agrare @gmcculloug @abellotti who else might be interested?

gmcculloug:
@Ladas Seeing @gtanzillo missing from the list is making me wonder if you meant him and not me.  🤔

Ladas:
@gmcculloug ah I wanted to add all gregs :-) @Fryguy said that your team should be involved too, since we want the request validation in all services

Ladas:
Implemented as `before_action`, @gmcculloug you should be able to use this in your projects for validation of API requests.

PR #79, opened 2019-06-12
Removed OpenAPI Regexp validation

slemrmartin:
Swagger doc Definition doesn't want '/' at the start/end of pattern.
https://swagger.io/docs/specification/data-models/data-types/#pattern

`OpenAPIParser` validates pattern and wraps pattern with '/', so this check is not compatible.
Wrong definition is for example there:
https://github.com/ManageIQ/sources-api/blob/master/public/doc/openapi-3-v1.0.0.json#L1200
check value against pattern here: https://github.com/ota42y/openapi_parser/blob/master/lib/openapi_parser/schema_validators/string_validator.rb#L50

Happened during implementation of https://github.com/ManageIQ/manageiq-api-common/issues/67

cc @bdunne @Ladas @abellotti

### DIFF
--- a/lib/manageiq/api/common/open_api/docs/component_collection.rb
+++ b/lib/manageiq/api/common/open_api/docs/component_collection.rb
@@ -57,8 +57,7 @@ module ManageIQ
             end
 
             def regexp_from_pattern(pattern)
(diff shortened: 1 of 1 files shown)


slemrmartin:
validation strategy based on https://github.com/ManageIQ/topological_inventory-ingress_api/pull/47

PR #82, opened 2019-06-20
Fix constant problems

bdunne:
Occasionally we get an error about missing constant ManageIQ::Api::Docs
which should be actually be finding ::Api::Docs in lib/api/docs.rb in the
apps, but for some reason that isn't loaded yet.  Instead, move all of this
logic to manageiq-api-common and create
::ManageIQ::API::Common::OpenApi::Docs.instance


bdunne:
cc @abellotti @syncrou

PR #83, opened 2019-06-21
Splitting the GraphQL OpenAPI description

abellotti:
- Moving the GraphQL Response into a separate schema definition as the client generator was getting confused with the inline response definition.


abellotti:
@miq-bot assign @bdunne

bdunne:
Do consumers need to be updated to call the new method to get the response object?

PR #84, opened 2019-06-24
Updated how we find the openapi json in OpenApi::Docs.instance

abellotti:
Updated how we find the OpenAPI json in OpenApi::Docs.instance so this code also works with catalog api as they have the OpenAPI docs in a different location than topological inventory and sources.

### DIFF
--- a/lib/manageiq/api/common/open_api/docs.rb
+++ b/lib/manageiq/api/common/open_api/docs.rb
@@ -8,7 +8,7 @@ module ManageIQ
       module OpenApi
         class Docs
           def self.instance
-            @instance ||= new(Dir.glob(Rails.root.join("public", "doc", "openapi*.json")))
+            @instance ||= new(Dir.glob(Rails.root.join("public", "**", "openapi*.json")))
           end
 
           def initialize(glob)


abellotti:
ping @bdunne can you review please ? 🙇 

/cc @gmcculloug @syncrou this is what was causing the graphql failure for catalog api, couldn't find the OpenAPI json to initialize the GraphQL schema.

lib/manageiq/api/common/open_api/docs.rb
```diff
@@ -8,7 +8,7 @@ module Common
       module OpenApi
         class Docs
           def self.instance
-            @instance ||= new(Dir.glob(Rails.root.join("public", "doc", "openapi*.json")))
+            @instance ||= new(Dir.glob(Rails.root.join("public", "**", "openapi*.json")))
```
bdunne: Where are other projects storing this file?  I thought we copied the path from catalog

abellotti: catalog is in public/catalog/v1.0.0/  whereas topo and sources is just public/doc/

abellotti: Instead of the open ended **, the other option is to require all consumers to have their openapi docs in public/doc/, but not sure what other constraints this would impose.

syncrou: @bdunne @abellotti my understanding around our `/public/catalog/v1.0.0/` path was to come inline with an IPP doc.

Fryguy: I'm not sure who, but someone has run an audit against all of these, so there must be some standardized path defined.

Fryguy:
I would prefer we have a standardized path, then everyone follows that...whether its public/doc or public/<appname>, I don't have a preference, but the former is likely easier to hardcode.

lib/manageiq/api/common/open_api/docs.rb
gmcculloug: I imagine the confusion might have came about because of this wording:
> For consistency and to support automated testing of APIs, all applications MUST serve the current OpenAPI specification in JSON format on $PATH_PREFIX/$APP/v$VERSION/openapi.json.

However, all the APIs do serve the openapi.json on that path, but the file lives in a different location on disk.

Catalog and Approval use a versioned folder that contains the `openapi.json` file:
https://github.com/ManageIQ/approval-api/tree/master/public/approval/v1.0

Topology adds versioning to the openapi.json filename  (`openapi-3-v#.#.#.json`): 
https://github.com/ManageIQ/topological_inventory-api/tree/master/public/doc

syncrou:
@Fryguy @abellotti @bdunne We moved our `openapi.json` into a `public/doc` folder and named it with the version in identical fashion with topological-api.

https://github.com/ManageIQ/catalog-api/pull/317

That PR is green, so I believe we can close this.

abellotti:
Closing this in preference to @syncrou 's PR https://github.com/ManageIQ/catalog-api/pull/317 for a common location for all.

PR #85, opened 2019-06-25
GraphQL Specs

abellotti:
- Updating the dummy application to include a subset of the sources database, schema, model, openapi spec, route and graphql controller.
- Include GraphQL Specs


abellotti:
[WIP]

- [ ] Depends on https://github.com/ManageIQ/manageiq-api-common/pull/78
- [ ] Add tests for
  - [ ] querying multiple primary collections
  - [ ] offset
  - [ ] limit
  - [ ] filter
  - [ ] querying associations
  - [ ] offset in association
  - [ ] limit in association

spec/dummy/app/models/concerns/password_concern.rb
```diff
@@ -0,0 +1,53 @@
+require "manageiq-password"
+
+module PasswordConcern
```
Fryguy: Don't duplicate PasswordConcern here.  It is already of the ManageIQ::Password gem and that module can just be used directly.

spec/dummy/app/models/authentication.rb
```diff
@@ -0,0 +1,7 @@
+class Authentication < ApplicationRecord
+  include PasswordConcern
```
Fryguy: Just use `ManageIQ::Password::PasswordMixin`, then no need to duplicate it in this project.

spec/dummy/app/models/authentication.rb
```diff
@@ -0,0 +1,7 @@
...
+  include TenancyConcern
+  encrypt_column :password
```
Fryguy: Minor, but it reads better to keep this line with the mixin inclusion since they are related.

spec/dummy/app/models/concerns/tenancy_concern.rb
```diff
@@ -0,0 +1,14 @@
+module TenancyConcern
```
Fryguy: Is this not already a mixin in this project?

spec/requests/api/graphql_spec.rb
```diff
@@ -0,0 +1,664 @@
...
+  let!(:source_typeR) { SourceType.create(:name => "rhev", :product_name => "RedHat Virtualization", :vendor => "redhat") }
+  let!(:source_typeV) { SourceType.create(:name => "vmware", :product_name => "VmWare vCenter", :vendor => "vmware") }
+
+  let!(:source_a1)    { Source.create!(:tenant => tenant, :uid => "1", :name => "source_a1", :source_type => source_typeR) }
+  let!(:source_a2)    { Source.create!(:tenant => tenant, :uid => "2", :name => "source_a2", :source_type => source_typeR) }
+  let!(:source_b1)    { Source.create!(:tenant => tenant, :uid => "3", :name => "source_b1", :source_type => source_typeR) }
+  let!(:source_b2)    { Source.create!(:tenant => tenant, :uid => "4", :name => "source_b2", :source_type => source_typeR) }
+  let!(:source_b3)    { Source.create!(:tenant => tenant, :uid => "5", :name => "source_b3", :source_type => source_typeR) }
+
+  let!(:endpoint_a21) { Endpoint.create!(:tenant => tenant, :source => source_a2, :host => "www.source_a2.com", :port => "121", :role => "web_lb1") }
```
Fryguy: Minor, but for the URLs, prefer subdomains of example.com (This way we don't accidentally point to "real" domains), so in this case `source_a2.example.com`

spec/requests/api/graphql_spec.rb
```diff
@@ -0,0 +1,664 @@
...
+          "sources": [
+            {
+              "name": "source_b1"
+            },
+            {
+              "name": "source_b2"
+            },
+            {
+              "name": "source_b3"
+            }
```
Fryguy: This feels like it may create sporadic test failures.  That is, unless you are sorting the names somehow, it's possible for them to come back in a different database order, and thus fail the eq matcher.

Fryguy:
@abellotti Can you rebase?

spec/dummy/app/models/concerns/tenancy_concern.rb
abellotti: not that I can see, the TenancyConcern here is specific to the sources api.

spec/requests/api/graphql_spec.rb
abellotti: well, PaginatedResponse sorts by id, and the expectation (sorry, couldn't help it 😉) is that the creates by the let!'s above will be done in order, and thus the returned names will be in that order too.  Am I missing some obscure rspec behavior that could bite us later ?

Fryguy:
can you rebase?

abellotti:
Just redid, Everything is up-to-date.

Fryguy:
ohhhh...sorry I saw the GraphQL controller and didn't realize it was added to the dummy app.

spec/dummy/app/controllers/api/graphql_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+
+module Api
+  class GraphqlController < ApplicationController
+    def query
+      graphql_api_schema = ::ManageIQ::API::Common::GraphQL::Generator.init_schema(request)
+      variables = ::ManageIQ::API::Common::GraphQL.ensure_hash(params[:variables])
+      result = graphql_api_schema.execute(
+        params[:query],
+        :variables => variables
+      )
```
Fryguy: It feels like we could encapsulate all of this into a simpler helper method such that this controller just becomes

```ruby
module Api
  class GraphqlController < ApplicationController
    def query
      render :json => ManageIQ::API::Common::GraphQL::ControllerAction.query(request)
    end
  end
end
```

and then separately

```ruby
module ManageIQ::API::Common::GraphQL::ControllerAction
  def self.query(request, params)
    GraphQL::Generator.init_schema(request).execute(
      params[:query],
      :variables => GraphQL.ensure_hash(params[:variables])
    )
  end
end
```

Fryguy:
LGTM :+1:

spec/dummy/app/controllers/api/graphql_controller.rb
abellotti: Do you want this done here or in a separate PR ?

abellotti: Separate PR it is.

abellotti:
@miq-bot assign @Fryguy

abellotti:
ping @Fryguy  can this be merged ?

abellotti:
@miq-bot assign @bdunne

bdunne:
It feels like there is a lot of extra schema in the dummy app that we don't need for these tests, but I'm okay with merging.

PR #86, opened 2019-07-05
Support Alternate OpenAPI Schema Class Name

abellotti:
Support for ClassName as well as ClassNameOut schema name definition in the OpenAPI spec.

This allows us support either the default ClassName or alternate ClassNameOut as used by the approval-api service. While a more elaborate implementation via the schema_overlay parameter can be done to allow for arbitrary name mappings, it seemed unnecessary for this one case (i.e. over engineered), we can revisit if another micro-service has a similar requirement but needing a different mapping.

### DIFF
--- a/lib/manageiq/api/common/graphql/generator.rb
+++ b/lib/manageiq/api/common/graphql/generator.rb
@@ -8,6 +8,14 @@ module ManageIQ
           PARAMETERS_PATH = "/components/parameters".freeze
           SCHEMAS_PATH = "/components/schemas".freeze
 
+          def self.openapi_schema(openapi_doc, klass_name)
+            schemas = openapi_doc.content.dig(*path_parts(SCHEMAS_PATH))
+            [klass_name, "#{klass_name}Out"].each do |name|
+              schema = schemas[name]
+              return [name, schema] if schema
+            end
+          end
+
           def self.path_parts(openapi_path)
             openapi_path.split("/")[1..-1]
           end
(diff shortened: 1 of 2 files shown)


abellotti:
ping @bzwei 

image

bzwei:
@abellotti it works for me. Thanks

abellotti:
ping @bdunne can you look over this PR when you get a chance ?

Needed for enabling GraphQL with approval-api micro-service.

Ran specs from https://github.com/ManageIQ/manageiq-api-common/pull/85 against it, and Bill's PR specs of course.

bdunne:
@abellotti I don't think I understand the problem that we're trying to solve here.  Can you give an example of what `ClassNameOut` is?

bzwei:
@bdunne Some services such as Rbac or Approval-api append `In` or `Out` to the model for input and output. For example for resource model `Workflow`, the output in openapi spec is `WorflowOut`.

bdunne:
@bzwei In an effort to maintain consistency between all of the APIs we're working on, is there a use case for having different schemas for the same thing?  Do they ever return totally different objects?  In the few cases that I quickly looked at, it was just a few extra fields on the `Out` object which look like they should be marked as read-only.

bdunne:
@bzwei ^

bzwei:
@bdunne in our early revisions the workflow model's input and output were very different. We learned from rbac service's naming convention by using suffix `In` and `Out`. We like it because of the clear meaning from the model name. Later we simplified workflow. Now its input and output are more compatible. It is too much if we go back to rename every model. Besides we still like the naming convention, not only it is meaningful but also it is flexible.

agrare:
@bzwei can you provide some examples where this would be useful?  I'm still failing to see how this is helpful, if the two are different then it sounds like they shouldn't be the same type of object...

Fryguy:
I am also failing to understand this.  OpenAPI has support for "read-only" fields specifically for this in/out use case.

Fryguy:
Even if I accepted that there are different input and output objects, hardcoding a naming convention feels wrong.  I would expect the OpenAPI spec to describe the "Out" object as the return type and then the graphql generator would just use that directly.  In fact, since we don't have mutators the "In" object would effectively never be referenced.

bzwei:
There is no openapi spec requirement that the schema name must match the resource name. In fact every resource has schema section in each of request and response body. I agree that we'd better not to hardcode any naming convention. GraphQL generator can follow the schema reference to get the model name, for example https://github.com/ManageIQ/approval-api/blob/master/public/doc/openapi-3-v1.0.0.json#L105.

PR #89, opened 2019-09-09
Share request body validation

bdunne:
Extracted from sources-api ApplicationController
Related to TPINVTRY-186

### DIFF
--- a/lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
+++ b/lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
@@ -2,34 +2,35 @@ module ManageIQ
   module API
     module Common
       module ApplicationControllerMixins
-        module InputValidation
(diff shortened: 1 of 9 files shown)


lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
```diff
@@ -0,0 +1,58 @@
...
+              render :json => error_document.to_h, :status => error_document.status
+            end
+          end
+
+          private
+
+          def body_params
+            @body_params ||= begin
+              raw_body       = request.body.read
+              parsed_body    = JSON.parse(raw_body)
```
agrare: Some extra padding here

bdunne: :+1: Fixed

PR #90, opened 2019-09-09
Adding support for an optional base_query 

abellotti:
Adding support for an optional base_query provided by the GraphQL controller for defining a base_query via the schema overlay for the different model queries.

Example would be something like:

```
  schema_overlay = {
    "^.*$" => {
      "base_query" => ::Api::GraphqlController.method(:base_query)
    }
  }

  graphql_api_schema = ::ManageIQ::API::Common::GraphQL::Generator.init_schema(request, schema_overlay)
```

Calls to ::Api::GraphqlController base_query would be as follows:

```
::Api::GraqhqlController.base_query(model_class, ctx)
```

Where ctx is the field resolver context.

### DIFF
--- a/lib/manageiq/api/common/graphql/generator.rb
+++ b/lib/manageiq/api/common/graphql/generator.rb
@@ -72,7 +72,7 @@ module ManageIQ
             schema_overlay.keys.each do |collection_regex|
               next unless collection.match(collection_regex)
 
-              collection_schema_overlay.merge!(schema_overlay.fetch_path(collection_regex) || {})
+              collection_schema_overlay.merge!(schema_overlay[collection_regex] || {})
             end
             collection_schema_overlay
           end
--- a/lib/manageiq/api/common/graphql/generator.rb
+++ b/lib/manageiq/api/common/graphql/generator.rb
@@ -67,6 +67,16 @@ module ManageIQ
             field_resolvers
           end
 
+          def self.collection_schema_overlay(schema_overlay, collection)
+            collection_schema_overlay = {}
+            schema_overlay.keys.each do |collection_regex|
+              next unless collection.match(collection_regex)
+
+              collection_schema_overlay.merge!(schema_overlay.fetch_path(collection_regex) || {})
+            end
+            collection_schema_overlay
+          end
+
           def self.init_schema(request, schema_overlay = {})
             api_version       = ::ManageIQ::API::Common::GraphQL.version(request)
             version_namespace = "V#{api_version.tr('.', 'x')}"
--- a/lib/manageiq/api/common/graphql/generator.rb
+++ b/lib/manageiq/api/common/graphql/generator.rb
@@ -68,13 +68,11 @@ module ManageIQ
           end
 
           def self.collection_schema_overlay(schema_overlay, collection)
-            collection_schema_overlay = {}
-            schema_overlay.keys.each do |collection_regex|
+            schema_overlay.keys.each_with_object({}) do |collection_regex, collection_schema_overlay|
               next unless collection.match(collection_regex)
 
               collection_schema_overlay.merge!(schema_overlay[collection_regex] || {})
             end
-            collection_schema_overlay
           end
 
           def self.init_schema(request, schema_overlay = {})
--- a/lib/manageiq/api/common/graphql/generator.rb
+++ b/lib/manageiq/api/common/graphql/generator.rb
@@ -69,7 +69,7 @@ module ManageIQ
 
           def self.collection_schema_overlay(schema_overlay, collection)
             schema_overlay.keys.each_with_object({}) do |collection_regex, collection_schema_overlay|
-              next unless collection.match(collection_regex)
+              next unless collection.match?(collection_regex)
 
               collection_schema_overlay.merge!(schema_overlay[collection_regex] || {})
             end
--- a/lib/manageiq/api/common/graphql/templates/query_type.erb
+++ b/lib/manageiq/api/common/graphql/templates/query_type.erb
@@ -24,8 +24,7 @@ QueryType = ::GraphQL::ObjectType.define do
       argument :filter, ::ManageIQ::API::Common::GraphQL::Types::QueryFilter, "The Query Filter for querying the #{collection}"
       resolve lambda { |_obj, args, ctx|
         if base_query.present?
-          impl, method = base_query.split('.')
-          scope = impl.constantize.send(method, model_class, ctx)
+          scope = base_query.call(model_class, ctx)
         else
           scope = model_class
         end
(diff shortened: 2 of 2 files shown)


abellotti:
@mkanoor, @syncrou, @bzwei  there you go, for your tire-kicking pleasures.

Essentially for the catalog-api, you'd define a base_query method so you define your optional scope similar to what https://github.com/ManageIQ/catalog-api/blob/0ca9977b0c5d2fdf9b9202d5053d5f9172656c83/app/controllers/api/v1x0/mixins/index_mixin.rb#L5 does.  you won't need the field resolver ctx in your case.

While you can override the base_query per collection, the general match for all should work fine here,

i.e. in your app/controllers/api/graphql_controller.rb

```
  def query 
     schema_overlay = {
        "^.*$" => {
           "base_query" => "::Api::GraphqlController.base_query"
        }
      }

      graphql_api_schema = ::ManageIQ::API::Common::GraphQL::Generator.init_schema(request, schema_overlay)
    ...
  end

  def self.base_query(model_class, _ctx)
     relation = rbac_scope(model_class.all) if RBAC::Access.enabled?
     ...
     relation
  end

```


/cc @gtanzillo @gmcculloug @Fryguy

lib/manageiq/api/common/graphql/generator.rb
```diff
@@ -67,6 +67,16 @@ def self.collection_field_resolvers(schema_overlay, collection)
             field_resolvers
           end
 
+          def self.collection_schema_overlay(schema_overlay, collection)
+            collection_schema_overlay = {}
+            schema_overlay.keys.each do |collection_regex|
+              next unless collection.match(collection_regex)
+
+              collection_schema_overlay.merge!(schema_overlay.fetch_path(collection_regex) || {})
```
Fryguy: `fetch_path` with a single element can just be `[]` instead

lib/manageiq/api/common/graphql/generator.rb
```diff
@@ -67,6 +67,16 @@ def self.collection_field_resolvers(schema_overlay, collection)
             field_resolvers
           end
 
+          def self.collection_schema_overlay(schema_overlay, collection)
+            collection_schema_overlay = {}
+            schema_overlay.keys.each do |collection_regex|
+              next unless collection.match(collection_regex)
```
Fryguy: Is this an actual regex here or a string?

Fryguy:
I like this approach and allowing the schema overlay as a general thing might allow us to do other interesting things.  

> While you can override the base_query per collection

How would one do it on a per collection basis?  Also, if we want a collection to *not* use the general one (say, there's no RBAC for a particular endpoint), how would one do that?

lib/manageiq/api/common/graphql/generator.rb
Fryguy: Also, since you are not using the return value of the match, nor any magic like `$1`, prefer `.match?` here

Fryguy:
@abellotti One thing I don't like is specifying the method as a String, which then you have to parse...something feels wrong about that, and possibly prone to error.  I'm thinking that since all of this is Ruby code (i.e. not as part of a JSON/YAML definition or anything), perhaps it's better to allow the base_query to be an actual proc?   So instead, your example might look like...

```ruby
def query 
  schema_overlay = {
    "^.*$" => {
      "base_query" => ::Api::GraphqlController.method(base_query)
    }
  }

  graphql_api_schema = ::ManageIQ::API::Common::GraphQL::Generator.init_schema(request, schema_overlay)
  ...
end
```

or it could even be an anonymous proc at that point, if the caller wanted

```ruby
def query 
  schema_overlay = {
    "^.*$" => {
      "base_query" => ->(model_class, _ctx) do
        relation = rbac_scope(model_class.all) if RBAC::Access.enabled?::Api::GraphqlController.method(base_query)
        ...
        relation
      end
    }
  }

  graphql_api_schema = ::ManageIQ::API::Common::GraphQL::Generator.init_schema(request, schema_overlay)
  ...
end
```

lib/manageiq/api/common/graphql/generator.rb
```diff
@@ -67,6 +67,16 @@ def self.collection_field_resolvers(schema_overlay, collection)
             field_resolvers
           end
 
+          def self.collection_schema_overlay(schema_overlay, collection)
+            collection_schema_overlay = {}
+            schema_overlay.keys.each do |collection_regex|
```
Fryguy: minor simplification, but this can be 

```ruby
schema_overlay.keys.each_with_object({}) do |collection_regex, collection_schema_overlay|
```

and then you don't need to add the extra explicit `collection_schema_overlay` return at the end

lib/manageiq/api/common/graphql/generator.rb
abellotti: these are regex'es

abellotti:
With the regex patterns, you'd specify individual collections as such:

```
schema_overlay = {
  "^portfolios$" => {
    "base_query" => "::Api::V1x0::PortfoliosController.base_query"
  },
  "^portfolio_items$" => {
    "base_query" => "::Api::V1x0::PortfolioItemsController.base_query"
  },
  ...
}
```

Optionally, we could anchor the incoming keys if not anchored with  ^ & $, this would
allow for:

```
schema_overlay = {
  "portfolios" => {
    "base_query" => "::Api::V1x0::PortfoliosController.base_query"
  },
  "portfolio_items" => {
    "base_query" => "::Api::V1x0::PortfolioItemsController.base_query"
  },
  ...
}
```


No current way to exclude/override one's entry if matched by a more
general regex. while we could do an empty "" specification for that entry,
there's no guarantee of the order of the merges. Might need to be supported
with additional keywords, like use_default_base_query or something
like that.

Fryguy:
I think we can require the anchors for now, and change it later if it becomes clunky...better to keep one way of doing things than try to do all possibilities up front.

Fryguy:
>  there's no guarantee of the order of the merges

Actually there is...since the caller controls the schema_overlay, and Ruby hashes are ordered by insertion order, then the caller has full control.

abellotti:
ooh nice, if the hashes are ordered by insertion order, then the more specific can be last, and they can simply say "base_query" => nil or "" for the ones that want the default, and it will just work, since the present? check is there https://github.com/ManageIQ/manageiq-api-common/pull/90/files#diff-36f774fe8151e1ddea8f9d6b08a0603cR26

Fryguy:
@abellotti Thoughts on https://github.com/ManageIQ/manageiq-api-common/pull/90#issuecomment-529984302 ?

abellotti:
I think the second would work better here ...

    "base_query" => lambda do |model_class, _ctx|
        do or call something with model_class
    end

let me kick that tire ...

Fryguy:
they are equivalent...  `.method` returns a proc based on the method asked for

abellotti:
just tested and it works with both, method reference as well as embedding the lambda. yea. I'll push  up another commit and update the example usage above.

Fryguy:
@abellotti Are you able to write specs for this change?

PR #93, opened 2019-09-13
Allow for empty bodies

mkanoor:
Some of the post might not have a body, return an empty hash
if the body is empty

### DIFF
--- a/lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
+++ b/lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
@@ -27,7 +27,7 @@ module ManageIQ
           def body_params
             @body_params ||= begin
               raw_body    = request.body.read
-              parsed_body = JSON.parse(raw_body)
+              parsed_body = raw_body.blank? ? {} : JSON.parse(raw_body)
               ActionController::Parameters.new(parsed_body).permit!
             rescue JSON::ParserError
               raise ManageIQ::API::Common::ApplicationControllerMixins::RequestBodyValidation::BodyParseError
--- a/spec/requests/request_body_validation_spec.rb
+++ b/spec/requests/request_body_validation_spec.rb
@@ -7,8 +7,7 @@ RSpec.describe "ManageIQ::API::Common::ApplicationController Body", :type => :re
   it "empty body" do
     post("/api/v1.0/authentications", :headers => headers, :params => "")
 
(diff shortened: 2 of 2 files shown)


mkanoor:
@bdunne 
Not sure how to get it to fail by passing in other valid JSON strings the parsing works but it fails with
#<NoMethodError: undefined method `with_indifferent_access' for "1":String>
when we call 
ActionController::Parameters.new(parsed_body).permit!

I have tried the spec with 
```
post("/api/v1.0/authentications", :headers => headers, :params => "1".to_json)
```
It only hits that exception when I pass in an empty string which we are short circuiting now.

Fryguy:
@mkanoor What is the use case for empty bodies?  I can't think of one place where you would want to respond with nothing.

bdunne:
> @bdunne
> Not sure how to get it to fail by passing in other valid JSON strings the parsing works but it fails with
> #<NoMethodError: undefined method `with_indifferent_access' for "1":String>
> when we call
> ActionController::Parameters.new(parsed_body).permit!
> 
> I have tried the spec with
> 
> ```
> post("/api/v1.0/authentications", :headers => headers, :params => "1".to_json)
> ```
> 
> It only hits that exception when I pass in an empty string which we are short circuiting now.

Maybe something like: `post("/api/v1.0/authentications", :headers => headers, :params => "{\"a\":\"z}")

mkanoor:
@bdunne 
That gets trapped before it gets to us
ActionDispatch::Http::Parameters::ParseError:
in 
```
.gem/ruby/2.4.4/gems/actionpack-5.2.3/lib/action_dispatch/http/parameters.rb:117:in `rescue in parse_formatted_parameters'
```

mkanoor:
@bdunne Do we need our code isn'r RAILS already doing all the parsing for us here
https://github.com/rails/rails/blob/84496b85cbb9933b2a2638b74ecf49c6295136c3/actionpack/lib/action_dispatch/http/parameters.rb#L50

Fryguy:
Since Rails is the one catching in valid JSON values (when Content-Type: application/json is passed), then we don't really need to test that pat, technically, but we may want to separately create a PR that ensures that path returns a 400 Bad Request error.  Since we also accept payloads where application/json is not set, I think that's the test we need to add here.

mkanoor:
@Fryguy @bdunne 
Added a test for invalid text data

PR #94, opened 2019-09-13
Support for non openapi controllers

mkanoor:
Only openapi specific controllers should get validated with the
openapi json spec.

This uses a class variable at the controller level 
openapi_enabled
to check if we the class supports openapi validation.

For the classes that dont need openapi_validation the developer can set it to false.

### DIFF
--- a/lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
+++ b/lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
@@ -38,6 +38,7 @@ module ManageIQ
           # - only for HTTP POST/PATCH
           def validate_request
             return unless request.post? || request.patch?
+            return unless self.class.to_s.split('::').length > 2
 
             api_version = self.class.send(:api_version)[1..-1].sub(/x/, ".")
 
--- a/lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
+++ b/lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
@@ -7,7 +7,6 @@ module ManageIQ
           end
 
           def self.included(other)
-            other.class_attribute :openapi_enabled, :default => true
             ActionController::Parameters.action_on_unpermitted_parameters = :raise
 
             other.before_action(:validate_request)
--- a/lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
+++ b/lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
@@ -7,7 +7,7 @@ module ManageIQ
           end
 
           def self.included(other)
-            @@openapi_enabled = true
+            other.class_attribute :openapi_enabled, :default => true
             ActionController::Parameters.action_on_unpermitted_parameters = :raise
 
             other.before_action(:validate_request)
@@ -39,7 +39,7 @@ module ManageIQ
           # - only for HTTP POST/PATCH
           def validate_request
             return unless request.post? || request.patch?
-            return unless @@openapi_enabled
+            return unless self.class.openapi_enabled
 
             api_version = self.class.send(:api_version)[1..-1].sub(/x/, ".")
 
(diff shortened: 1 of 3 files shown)


lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
```diff
@@ -38,6 +38,7 @@ def body_params
           # - only for HTTP POST/PATCH
           def validate_request
             return unless request.post? || request.patch?
+            return unless self.class.to_s.split('::').length > 2
```
bdunne: Is there a better way to do this?  Maybe looking at the request path and only validating if the path is in the OpenAPI document?

bdunne: Out of curiosity, do you have a use case for undocumented controllers?

mkanoor: @bdunne We need the version and the path to get to the openapi document. This call fails before that.

The use case that we have is to limit the number of rescue_froms that we need to write in the controllers.
Erik was following this https://coderwall.com/p/iumbag/quickly-enable-middleware-based-custom-exception-handling-in-rails as an alternative to using rescue_froms. We have a controller call ErrorsController which is not path'ed for the openapi.

https://github.com/ManageIQ/catalog-api/blob/master/app/controllers/errors_controller.rb

Which emanates from the routes defined here
https://github.com/ManageIQ/catalog-api/blob/1ade13072b252848ace0e12e0312addb8fcdc933/config/routes.rb#L7

Fryguy: I don't understand how this change means "non openapi controller".  I read it as "any controller with a namespace", which isn't the same thing.

mkanoor: @Fryguy Can you please review changed it to use a class variable that can be set in non openapi controllers.

Fryguy:
Please use a class instance var...class var affects the base class as well.

mkanoor:
@Fryguy Please review

lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
```diff
@@ -7,6 +7,7 @@ class BodyParseError < ::RuntimeError
           end
 
           def self.included(other)
+            other.class_attribute :openapi_enabled, :default => true
```
Fryguy: LGTM though I'm not sure that this parameter belongs in this particular module?  Not sure what other module we would use though.  @bdunne Thoughts?

mkanoor:
@Fryguy Based on our discussion I have moved the class attribute into a mixin. Once this PR is merged I will include it in the sources API here
https://github.com/ManageIQ/sources-api/blob/c87a3e94f580e39db3305e6c889b8d47a4c04303/app/controllers/application_controller.rb#L6

lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
```diff
@@ -38,6 +38,7 @@ def body_params
           # - only for HTTP POST/PATCH
           def validate_request
             return unless request.post? || request.patch?
+            return unless self.class.openapi_enabled
```
Fryguy: Since this module depends on the other one, I *think* you can include it in this module so it's brought in automatically.  At present, if I merge this is will break other apps because of the dependency.

lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb
```diff
@@ -8,6 +8,7 @@ class BodyParseError < ::RuntimeError
 
           def self.included(other)
             ActionController::Parameters.action_on_unpermitted_parameters = :raise
+            other.include(OpenapiEnabled)
```
Fryguy: Super minor but can you add a blank line above this...otherwise it looks like this line is related to the previous line.

PR #97, opened 2019-09-18
Validate IDs when provided in the request path

bdunne:
If invalid, return a 400 error response
Fixes TPINVTRY-675
Fixes TPINVTRY-676

### DIFF
--- a/lib/manageiq/api/common/application_controller_mixins/api_doc.rb
+++ b/lib/manageiq/api/common/application_controller_mixins/api_doc.rb
@@ -9,6 +9,10 @@ module ManageIQ
 
           private
 
+          def api_doc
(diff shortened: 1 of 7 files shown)


agrare:
@bdunne this looks like it works for subcollections like

```
curl http://localhost:3000/api/v1.0/source_types/abcd/sources
{"errors":[{"status":400,"detail":"ID is invalid"}]}
```

But the primary collection still returns a 404:
```
http://localhost:3000/api/v1.0/source_types/abcd
{"errors":[{"status":404,"detail":"Record not found"}]}
```

bdunne:
Ok @agrare I moved the validation to a `before_action`, no changes necessary on the consumer side

PR #99, opened 2019-09-19
Parameter mixin

mkanoor:
Moved code from sources-api into the common gem for strong parameter check based on openapi.json

The original code was here
https://github.com/ManageIQ/sources-api/blob/0c096f39ee826b8269e79a28bc01475183ae2475/app/controllers/application_controller.rb#L74

### DIFF
--- a/lib/manageiq/api/common/application_controller_mixins/parameters.rb
+++ b/lib/manageiq/api/common/application_controller_mixins/parameters.rb
@@ -1,0 +1,95 @@
+module ManageIQ
+  module API
+    module Common
+      module ApplicationControllerMixins
+        module Parameters
+          def params_for_create
+            # We already validate this with OpenAPI validator, that validates every request, so we shouldn't do it again here.
+            body_params.permit!
+          end
+
+          def safe_params_for_list
+            # :limit & :offset can be passed in for pagination purposes, but shouldn't show up as params for filtering purposes
+            @safe_params_for_list ||= params.merge(params_for_polymorphic_subcollection).permit(*permitted_params, :filter => {})
+          end
+
+          def permitted_params
+            api_doc_definition.all_attributes + [:limit, :offset] + [subcollection_foreign_key]
+          end
+
+          def subcollection_foreign_key
+            "#{request_path_parts["primary_collection_name"].singularize}_id"
+          end
+
+          def params_for_polymorphic_subcollection
+            return {} unless subcollection?
+            return {} unless reflection = primary_collection_model&.reflect_on_association(request_path_parts["subcollection_name"])
+            return {} unless as = reflection.options[:as]
+            {"#{as}_type" => primary_collection_model.name, "#{as}_id" => request_path_parts["primary_collection_id"]}
+          end
+
+          def primary_collection_model
+            @primary_collection_model ||= request_path_parts["primary_collection_name"].singularize.classify.safe_constantize
+          end
+
+          def params_for_list
+            safe_params = safe_params_for_list.slice(*all_attributes_for_index)
+            if safe_params[subcollection_foreign_key_using_through_relation]
+              # If this is a through relation, we need to replace the :foreign_key by the foreign key with right table
+              # information. So e.g. :container_images with :tags subcollection will have {:container_image_id => ID} and we need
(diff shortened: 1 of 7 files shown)


mkanoor:
@bdunne @Fryguy 
Please review

lib/manageiq/api/common/application_controller_mixins/parameters.rb
```diff
@@ -0,0 +1,95 @@
+module ManageIQ
+  module API
+    module Common
+      module ApplicationControllerMixins
+        module Parameters
+          def params_for_create
+            # We already validate this with OpenAPI validator, that validates every request, so we shouldn't do it again here.
+            body_params.permit!
```
Fryguy: Now that you have openapi_enabled = false...it's possible we could now be permitting bad params to your non-openapi_enabled controllers... cc @bdunne

Fryguy: This probably goes for almost all of the methods in here.

spec/dummy/app/controllers/api/v1x0.rb
```diff
@@ -21,5 +21,22 @@ def show
         render :json => {:id => request_path_parts["primary_collection_id"]}.to_json
       end
     end
+
+    class PersonsController < ApplicationController
+      def create
+        params_for_create
```
Fryguy: What does this test besides just running the methods?

lib/manageiq/api/common/application_controller_mixins/parameters.rb
```diff
@@ -0,0 +1,95 @@
...
+            # :limit & :offset can be passed in for pagination purposes, but shouldn't show up as params for filtering purposes
+            @safe_params_for_list ||= params.merge(params_for_polymorphic_subcollection).permit(*permitted_params, :filter => {})
+          end
+
+          def permitted_params
+            api_doc_definition.all_attributes + [:limit, :offset] + [subcollection_foreign_key]
+          end
+
+          def subcollection_foreign_key
+            "#{request_path_parts["primary_collection_name"].singularize}_id"
```
Fryguy: I know it's not from this PR, but we a more Rails oriented way to do this...there's should be a helper that gets you this string

lib/manageiq/api/common/application_controller_mixins/parameters.rb
mkanoor: @Fryguy But you would be calling these methods only from openapi enabled controllers. A non openapi enabled controller doesn't have to call these methods

lib/manageiq/api/common/application_controller_mixins/parameters.rb
```diff
@@ -0,0 +1,95 @@
...
+          end
+
+          def params_for_polymorphic_subcollection
+            return {} unless subcollection?
+            return {} unless reflection = primary_collection_model&.reflect_on_association(request_path_parts["subcollection_name"])
+            return {} unless as = reflection.options[:as]
+            {"#{as}_type" => primary_collection_model.name, "#{as}_id" => request_path_parts["primary_collection_id"]}
```
Fryguy: same here - I know it's not from this PR, but we a more Rails oriented way to do this...there's should be a helper that gets you this string  cc @bdunne

spec/dummy/app/controllers/api/v1x0.rb
mkanoor: @Fryguy It just calls the strong parameters check and if its bad we get a 400 back.

lib/manageiq/api/common/application_controller_mixins/parameters.rb
Fryguy: No, but it's not uncommon from a controller to call methods like `params_for_create`, which would now be available to them.  At a minimum maybe these methods should raise an Exception unless openapi_enabled?

spec/dummy/app/controllers/api/v1x0.rb
Fryguy: Still feels like there should be a render call or something...Standalone it doesn't feel like a realistic looking tests.

mkanoor: @Fryguy Please review I have added a check for openapi.

PR #100, opened 2019-09-23
Default to add ':as => :json' for all patch, post, and put request specs

eclarizio:
Opening this PR in favor of https://github.com/ManageIQ/catalog-api/pull/369, since it appears that more than just catalog will make use of this spec change.


mkanoor:
@eclarizio 
Can you fix these specs so that the DefaultAsJson helps
https://github.com/ManageIQ/manageiq-api-common/blob/a2f413e12ce7b8596bcb8c25922c49a8145b6d22/spec/requests/request_body_validation_spec.rb#L2

bdunne:
It's fairly common for the GraphQL servers to support POST with application/graphql.  Is that a thing we want to allow?

mkanoor:
@bdunne 
isn't ours a REST API end point rather than a Graphql  end point. I am not sure if the generated clients would know how to  build an application/graphql content type.

PR #105, opened 2019-09-25
Adding GraphQL::Generator tests

### LINKED ISSUE #92: Need specs for GraphQL base_query
Follow up to #90 

cc @abellotti

abellotti:
Adding GraphQL::Generator test for the schema_overlay logic to test base_query

Fixes #92

### DIFF
--- a/lib/manageiq/api/common/graphql.rb
+++ b/lib/manageiq/api/common/graphql.rb
@@ -9,14 +9,12 @@ require "manageiq/api/common/graphql/types/big_int"
 require "manageiq/api/common/graphql/types/date_time"
 require "manageiq/api/common/graphql/types/query_filter"
 
-module ManageIQGraphQLApi
-end
-
 module ManageIQ
   module API
     module Common
       module GraphQL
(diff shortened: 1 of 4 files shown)


abellotti:
Keeping as WIP for now, depending on order, other tests unreliability fail.

abellotti:
The random spec failures were constant problems.  Moved the dynamically generated GraphQL schema away from ::Api into its own ::ManageIQGraphQLApi.

lib/manageiq/api/common/graphql.rb
```diff
@@ -9,10 +9,15 @@
 require "manageiq/api/common/graphql/types/date_time"
 require "manageiq/api/common/graphql/types/query_filter"
 
+module ManageIQGraphQLApi
```
bdunne: I don't like that this constant is not in the `ManageIQ::Api::Common` namespace

abellotti:
Does this fit the bill ?

```
::ManageIQ::API::Common::GraphQL::Api
```

PR #106, opened 2019-09-25
Moved RBAC logics from Catalog/Approval into api-common gem

hsong-rh:


abellotti:
I wonder if it shouldn't go in its own gem, i.e. manageiq-api-rbac. I may be wrong. @Fryguy thoughts ?

mkanoor:
@abellotti There is already a rbac client gem this is just wrappers and helper functions that we share between projects.
https://github.com/RedHatInsights/insights-rbac-api-client-ruby
Thats a gem that is the core RBAC logic.

Fryguy:
@bdunne Can you review as well?

Gemfile
```diff
@@ -6,4 +6,11 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 # Declare your gem's dependencies in topological_inventory.gemspec.
 # Bundler will treat runtime dependencies like base dependencies, and
 # development dependencies will be added by default to the :development group.
+gem 'rbac-api-client', :git => "https://github.com/RedHatInsights/insights-rbac-api-client-ruby", :branch => "master"
+
+group :development, :test do
+  gem 'climate_control'
```
Fryguy: I don't understand the point of adding this gem.  Would rather we could avoid this.

lib/manageiq/api/common/rbac/access.rb
```diff
@@ -0,0 +1,72 @@
...
+            @resource = resource
+            @verb     = verb
+            @regexp   = Regexp.new(":(#{@resource}|\\*):(#{@verb}|\\*)")
+            @app_name = ENV["APP_NAME"]
+          end
+
+          def process
+            ManageIQ::API::Common::RBAC::Service.call(RBACApiClient::AccessApi) do |api|
+              Rails.logger.info("Fetch access list for #{@app_name}")
+              @acl = ManageIQ::API::Common::RBAC::Service.paginate(api, :get_principal_access, {:limit => DEFAULT_LIMIT}, @app_name).select do |item|
```
Fryguy: You shouldn't need to keep specifying `ManageIQ::API::Common::RBAC`.  `Service.paginate` should be sufficient and would trim up the readability a bit.

lib/manageiq/api/common/rbac/access.rb
```diff
@@ -0,0 +1,72 @@
...
+  module API
+    module Common
+      module RBAC
+        class Access
+          attr_reader :acl
+          DEFAULT_LIMIT = 500
+          def initialize(resource, verb)
+            @resource = resource
+            @verb     = verb
+            @regexp   = Regexp.new(":(#{@resource}|\\*):(#{@verb}|\\*)")
```
Fryguy: This feels like the wrong way to build this regex, particularly if resource or verb have regexp characters in them (particularly a `.` or a paren)...I think this is more accurate:

```ruby
@regexp   = /:(#{Regexp.escape(@resource)}|\\*):(#{Regexp.escape(@verb)}|\\*)/
```

lib/manageiq/api/common/rbac/access.rb
```diff
@@ -0,0 +1,72 @@
...
+              @acl = ManageIQ::API::Common::RBAC::Service.paginate(api, :get_principal_access, {:limit => DEFAULT_LIMIT}, @app_name).select do |item|
+                Rails.logger.info("Found ACL: #{item}")
+                @regexp.match(item.permission)
+              end
+            end
+            self
+          end
+
+          def accessible?
+            Rails.logger.info("ACL for #{@app_name} #{@acl}")
```
Fryguy: This log message is strange as it just lists the ACLs but every time you call this method?

lib/manageiq/api/common/rbac/access.rb
```diff
@@ -0,0 +1,72 @@
...
+            @ids.include?('*') ? false : owner_scope_filter?
+          end
+
+          def self.enabled?
+            ENV['BYPASS_RBAC'].blank?
+          end
+
+          private
+
+          def generate_ids
```
Fryguy: Prefer just having an `ids` method that memoizes.  Then you don't need a separate generate method and then an `@ids` accessor.

Gemfile
hsong-rh: This gem is used in rspec file to set env variable for testing

lib/manageiq/api/common/rbac/access.rb
hsong-rh: Removed

Gemfile
Fryguy: right, but you can also just use existing RSpec stuff to do that.

lib/manageiq/api/common/rbac/access.rb
```diff
@@ -0,0 +1,68 @@
...
+              end
+            end
+          end
+
+          def owner_scope_filter?
+            @acl.any? do |item|
+              item.resource_definitions.any? do |rd|
+                rd.attribute_filter.key == 'owner' &&
+                  rd.attribute_filter.operation == 'equal' &&
+                  rd.attribute_filter.value == '{{username}}'
```
Fryguy: Is this literally `{{username}}` or should this have been a parameter?

lib/manageiq/api/common/rbac/access.rb
```diff
@@ -0,0 +1,68 @@
...
+            @verb     = verb
+            @regexp   = Regexp.new(":(#{Regexp.escape(@resource)}|\\*):(#{Regexp.escape(@verb)}|\\*)")
+            @app_name = ENV["APP_NAME"]
+          end
+
+          def process
+            Service.call(RBACApiClient::AccessApi) do |api|
+              @acl = Service.paginate(api, :get_principal_access, {:limit => DEFAULT_LIMIT}, @app_name).select do |item|
+                Rails.logger.info("Found ACL: #{item}")
+                @regexp.match(item.permission)
```
Fryguy: Since you are not using the return value from the match, prefer `.match?`

lib/manageiq/api/common/rbac/access.rb
```diff
@@ -0,0 +1,68 @@
...
+            @resource = resource
+            @verb     = verb
+            @regexp   = Regexp.new(":(#{Regexp.escape(@resource)}|\\*):(#{Regexp.escape(@verb)}|\\*)")
+            @app_name = ENV["APP_NAME"]
+          end
+
+          def process
+            Service.call(RBACApiClient::AccessApi) do |api|
+              @acl = Service.paginate(api, :get_principal_access, {:limit => DEFAULT_LIMIT}, @app_name).select do |item|
+                Rails.logger.info("Found ACL: #{item}")
```
Fryguy: This will get written out for every ACL.  Is that intentional?  a) that seems like a lot of logging and b) it was also print for the ACLs that don't match.

lib/manageiq/api/common/rbac/access.rb
```diff
@@ -0,0 +1,68 @@
...
+                next unless rd.attribute_filter.key == 'id'
+                next unless rd.attribute_filter.operation == 'equal'
+
+                ids << rd.attribute_filter.value
+              end
+            end
+          end
+
+          def owner_scope_filter?
+            @acl.any? do |item|
```
Fryguy: Will `@acl` change?  If not, this might be better memoized, since this feels like it could be relatively expensive.

lib/manageiq/api/common/rbac/acls.rb
```diff
@@ -0,0 +1,74 @@
+module ManageIQ
+  module API
+    module Common
+      module RBAC
+        class ACLS
```
Fryguy: Minor, but should this be `ACLs` (lower case s)`?

lib/manageiq/api/common/rbac/acls.rb
```diff
@@ -0,0 +1,74 @@
...
+          def create(resource_id, permissions)
+            permissions.collect do |permission|
+              create_acl(permission, resource_id)
+            end
+          end
+
+          def remove(acls, resource_id, permissions)
```
Fryguy: It feels weird to me that acls is a parameter when you have a class that represents ACLs.  That is, the ACLs themselves should be part of the class and the methods act on it.  Or perhaps what would be nicer is an ACLCollection class that itself has the remove, add, etc. methods on it, with create returning an ACLCollection?

Fryguy: (Doesn't have to be for this PR, but we might want to consider reorganizing as it's not very OO in design at the moment)

lib/manageiq/api/common/rbac/roles.rb
```diff
@@ -0,0 +1,77 @@
...
+          end
+
+          def find(name)
+            uuid = @roles[name]
+            get(uuid) if uuid
+          end
+
+          def with_each_role
+            @roles.each_value do |uuid|
+              yield get(uuid)
```
Fryguy: yikes...we hit the API in a loop?

lib/manageiq/api/common/rbac/roles.rb
```diff
@@ -0,0 +1,77 @@
...
+              role_in.name = name
+              role_in.access = acls
+              api_instance.create_roles(role_in).tap do |role|
+                @roles[name] = role.uuid
+              end
+            end
+          end
+
+          def update(role)
+            Service.call(RBACApiClient::RoleApi) do |api_instance|
```
Fryguy: Should this method be modifying `@roles` somewhere?

lib/manageiq/api/common/rbac/service.rb
```diff
@@ -0,0 +1,66 @@
+require 'rbac-api-client'
```
Fryguy: Also move down...maybe into .call?

lib/manageiq/api/common/rbac/service.rb
```diff
@@ -0,0 +1,66 @@
...
+        class Service
+          def self.call(klass)
+            setup
+            yield init(klass)
+          rescue RBACApiClient::ApiError => err
+            Rails.logger.error("RBACApiClient::ApiError #{err.message} ")
+            raise
+          end
+
+          def self.paginate(obj, method, pagination_options, *method_args)
```
Fryguy: This feels like an even higher-level method that just for RBAC Service as everything needs to paginate.  cc @bdunne

lib/manageiq/api/common/rbac/service.rb
```diff
@@ -0,0 +1,66 @@
...
+            Rails.logger.info("Sending Headers to RBAC #{headers}")
+            klass.new.tap do |api|
+              api.api_client.default_headers = api.api_client.default_headers.merge(headers)
+            end
+          end
+
+          private_class_method def self.dev_credentials(config)
+            # Set up user/pass for basic auth if we're in dev and they exist.
+            if Rails.env.development?
+              config.username = ENV['DEV_USERNAME'] || raise("Empty ENV variable: DEV_USERNAME")
```
Fryguy: You can use `ENV.fetch('DEV_USERNAME')` and it will automatically raise on a missing value so no need to do it manually.

lib/manageiq/api/common/rbac/share_resource.rb
```diff
@@ -0,0 +1,60 @@
+# load the gem
+require 'rbac-api-client'
```
Fryguy: Ditto

lib/manageiq/api/common/rbac/query_shared_resource.rb
```diff
@@ -0,0 +1,44 @@
+require 'rbac-api-client'
```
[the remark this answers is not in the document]
Fryguy: Looking back maybe this is ok.  How are these files loaded by the caller?  Do they just require the base rbac file, and then the rest load?  If so, that effectively delay loads and then it's probably ok.

lib/manageiq/api/common/rbac/share_resource.rb
```diff
@@ -0,0 +1,60 @@
+# load the gem
```
Fryguy: Unnecessary comment.

lib/manageiq/api/common/rbac/unshare_resource.rb
```diff
@@ -0,0 +1,30 @@
+require 'rbac-api-client'
+module ManageIQ
+  module API
+    module Common
+      module RBAC
+        class UnshareResource < ShareResource
```
Fryguy: What is an unshare resource? and how can an unshare be a subclass of a share?

lib/manageiq/api/common/rbac/unshare_resource.rb
```diff
@@ -0,0 +1,30 @@
...
+          attr_accessor :count
```
Fryguy: Minor but blank line after the attr_accessors as they are not a "part" of initialize

spec/support/service_spec_helper.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+  end
+
+  def default_headers
+    { 'x-rh-identity'            => encoded_user_hash,
+      'x-rh-insights-request-id' => 'gobbledygook' }
+  end
+
+  def original_url
+    "http://whatever.com"
```
Fryguy: Prefer example.com as it's an actual URL that's designed for testing setups.

lib/manageiq/api/common/rbac/service.rb
bdunne: It feels like something that belongs in an API client

bdunne: Would `method` ever not be `:get`?

lib/manageiq/api/common/rbac/access.rb
hsong-rh: yes, it's literal

lib/manageiq/api/common/rbac/acls.rb
hsong-rh: Will change in the future if needs

lib/manageiq/api/common/rbac/roles.rb
hsong-rh: We only keep name and uuid in `@roles`

lib/manageiq/api/common/rbac/roles.rb
hsong-rh: The `@roles` here already filtered by `prefix` and current user.

lib/manageiq/api/common/rbac/service.rb
hsong-rh: API client doesn't support pagination, so we add it here.

lib/manageiq/api/common/rbac/unshare_resource.rb
hsong-rh: The `share` and `unshare` resource are different, but they share the similar codes.

lib/manageiq/api/common/rbac/access.rb
hsong-rh: Removed

lib/manageiq/api/common/rbac/acls.rb
hsong-rh: Changed into `ACL`

Gemfile
```diff
@@ -6,4 +6,10 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 # Declare your gem's dependencies in topological_inventory.gemspec.
 # Bundler will treat runtime dependencies like base dependencies, and
 # development dependencies will be added by default to the :development group.
+gem 'rbac-api-client', :git => "https://github.com/RedHatInsights/insights-rbac-api-client-ruby", :branch => "master"
+
+group :development, :test do
+  gem 'webmock'
```
Fryguy: Should this be in the gemspec as a development dependency?

lib/manageiq/api/common/rbac/access.rb
Fryguy: @hsong-rh I unresolved this comment because you didn't change it...or did you choose not to?

lib/manageiq/api/common/rbac/acl.rb
```diff
@@ -0,0 +1,74 @@
...
+            end
+          end
+
+          def add(acls, resource_id, permissions)
+            new_acls = permissions.each_with_object([]) do |permission, as|
+              next if find_matching(acls, resource_id, permission)
+
+              as << create_acl(permission, resource_id)
+            end
+            new_acls.empty? ? acls : new_acls + acls
```
Fryguy: If it's empty, then the `+` operation will work as expected, so this line can just be `new_acls + acls`

Fryguy:
Out of curiosity...why do we have to do some much fine-grained checking on the client side?  A lot of this code is looping over specific permissions and checking arrays for content.  I would have expected an API call to rbac asking "for this user and this set of permissions, which are they allowed or not allowed to do".  Then it's a single API call and all of this logic lives on the other side of the fence within the microservice.

mkanoor:
@Fryguy the RBAC service provides a database for groups, roles and policies. Its uptown each app on how they use these objects. In our case we have to share portfolios with specific permissions (read, write, order) with specific groups. So the same portfolio could be shared with different groups with different permissions. This information is stored in the resource definition. in the roles object, so we have to traverse these objects to figure out the net permission for each user at runtime. There isn't an Resource based API in the RBAC service, sharing/unsharing is implemented by the Catalog Service. We have tried to optimize some of the calls by naming 'roles' with the portfolio_id and group_id so we can quickly get to them as opposed to going thru a whole resolution process on the RBAC service side.

Gemfile
Fryguy: If it's in the gemspec you should then remove this line?

lib/manageiq/api/common/rbac/access.rb
```diff
@@ -0,0 +1,66 @@
...
+          def initialize(resource, verb)
+            @resource = resource
+            @verb     = verb
+            @regexp   = Regexp.new(":(#{Regexp.escape(@resource)}|\\*):(#{Regexp.escape(@verb)}|\\*)")
+            @app_name = ENV["APP_NAME"]
+          end
+
+          def process
+            Service.call(RBACApiClient::AccessApi) do |api|
+              @acl ||= Service.paginate(api, :get_principal_access, {:limit => DEFAULT_LIMIT}, @app_name).select do |item|
```
hsong-rh: @Fryguy I memorized `@acl` here

lib/manageiq/api/common/rbac/access.rb
hsong-rh: I cached `@acl` at https://github.com/ManageIQ/manageiq-api-common/blob/71e33b021610fcb6d3d82f3758ee30ca8a29864b/lib/manageiq/api/common/rbac/access.rb#L17

Gemfile
hsong-rh: Removed.

hsong-rh:
@Fryguy anything else need to change?

Gemfile
```diff
@@ -6,4 +6,6 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 # Declare your gem's dependencies in topological_inventory.gemspec.
 # Bundler will treat runtime dependencies like base dependencies, and
 # development dependencies will be added by default to the :development group.
+gem 'rbac-api-client', :git => "https://github.com/RedHatInsights/insights-rbac-api-client-ruby", :branch => "master"
```
bdunne: Just a note that this will have to be in the consumers Gemfile as well...

PR #107, opened 2019-09-26
Enhanced GraphQL parsing of OpenAPI component paths

abellotti:
Enhanced the GraphQL schema generator to support other than the {id} references in the paths, namely {anything_id} with a more flexible {[[a-z]*_]*id} match

### DIFF
--- a/lib/manageiq/api/common/graphql/generator.rb
+++ b/lib/manageiq/api/common/graphql/generator.rb
@@ -41,16 +41,19 @@ module ManageIQ
 
           def self.resource_associations(openapi_content, collection)
             collection_is_associated = openapi_content["paths"].keys.any? do |path|
-              path.match("^/[^/]*/{id}/#{collection}$") &&
-                openapi_content.dig("paths", path, "get")
+              path.match?("^/[^/]*/{[[a-z]*_]*id}/#{collection}$") &&
+                openapi_content.dig("paths", path, "get").present?
             end
             collection_associations = []
             openapi_content["paths"].keys.each do |path|
-              subcollection_match = path.match("^/#{collection}/{id}/([^/]*)$")
+              subcollection_match = path.match("^/#{collection}/{[[a-z]*_]*id}/([^/]*)$")
               next unless subcollection_match
 
               subcollection = subcollection_match[1]
-              next unless openapi_content.dig("paths", "/#{subcollection}/{id}", "get")
+              next unless openapi_content["paths"].keys.any? do |subcollection_path|
+                subcollection_path.match?("^/#{subcollection}/{[[a-z]*_]*id}$") &&
+                openapi_content.dig("paths", subcollection_path, "get").present?
+              end
 
               collection_associations << subcollection
             end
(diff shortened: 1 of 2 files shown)


abellotti:
/cc @bzwei @mkanoor 

Bill, I think this was inhibiting the catalog and approval api's graphql schema to follow associations since components paths are there like /portfolios/{portfolio_id}/portfolio_items, so the original path checks of {id} were not matching.  Can you try this with your API and let me know if it now allows you to traverse associations with GraphiQL.   I'll unwip if it works for you.

bzwei:
@abellotti how do I use this? For example
```
  "query": "{workflows(id: 1) { name template_id }}"
```
works, but
```
  "query": "{workflows(id: 1) { name template }}"
```
still fails with error "Field 'template' doesn't exist on type 'Workflow'"

abellotti:
Hi @bzwei, I don't see template as an attribute or templates exposed as subcollection of workflows, just template_id which is listed in WorkFlowOut, that explains what you noticed.

In approval-api openapi, you have:
```
  "/workflows":
  "/workflows/{id}":
  "/workflows/{workflow_id}/requests":
```

So with GraphiQL, you should be able to see the requests association for your app,
so something like the following should work:

```
{
  workflows(id: 1) {
    name
    template_id
    requests {
      id
      name
      state
   }
 }
}
```

If the above works for you, that means this PR is working (ability to now traverse to the requests association as declared via the /workflows/{workflow_id}/requests component path in the OpenAPI spec.

bzwei:
@abellotti Yes. It works as you described. So we can only include subcollections but not expand belong_to association, right?

abellotti:
correct, associations as declared by the OpenAPI spec.

spec/dummy/public/doc/openapi-3-v1.0.0.json
```diff
@@ -263,7 +263,7 @@
         }
       }
     },
-    "/source_types/{id}/sources": {
```
bdunne: This no longer matches the documented parameter name

lib/manageiq/api/common/graphql/generator.rb
```diff
@@ -100,7 +103,7 @@ def self.init_schema(request, schema_overlay = {})
             resources.each do |resource|
               next unless openapi_content.dig("paths", resource, "get") # we only care for queries
 
-              rmatch = resource.match("^/(.*/)?([^/]*)/{id}$")
+              rmatch = resource.match("^/(.*/)?([^/]*)/{[[a-z]*_]*id}$")
```
bdunne: Can this be `/^/(.*/)?([^/]*)/{[a-z_]*id}$/`

bdunne: same in the other examples

abellotti: I had that for a bit, but it allows for prefixes with no _, so attrid vs requiring attr_id. I wanted either {id} or {..._id}.  wdyt ?

spec/dummy/public/doc/openapi-3-v1.0.0.json
abellotti: 👍 

- Updated the openapi json to:
   -  properly define the source_type_id
   -  remove entries not defined in the schema (or used in specs)
   -  fixed missing item type definition for the array field.
This openapi spec now passes the swagger editor validation.

PR #110, opened 2019-10-01
Updated the OpenAPI to declare the GraphQL request as a separate entry

abellotti:
Updated the OpenAPI auto-generation and docs have the GraphQL declared via a separate schema entry.


abellotti:
/cc @mkanoor can you try out with your client for catalog and let me know ?  Thanks.

I'll create the related PRs for topo and sources

mkanoor:
Looks good, previously it would go in as an Un-named object
👍

PR #113, opened 2019-10-01
Add new entitlements: Ansible and Migrations

gmcculloug:
You can see current entitlements with:
```
curl https://cloud.redhat.com/api/entitlements/v1/services -u user:password
```

Example output:
```
{"hybrid_cloud":{"is_entitled":true},"insights":{"is_entitled":true},"openshift":{"is_entitled":true},"smart_management":{"is_entitled":true},"ansible":{"is_entitled":false},"migrations":{"is_entitled":true}}
```

### DIFF
--- a/lib/manageiq/api/common/entitlement.rb
+++ b/lib/manageiq/api/common/entitlement.rb
@@ -9,8 +9,10 @@ module ManageIQ
         end
 
         %w[
+          ansible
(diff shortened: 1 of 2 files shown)


gmcculloug:
@Fryguy @bdunne Not sure who should review.

PR #114, opened 2019-10-02
Add ExceptionHandling mixin and move ApiError rescue_from

lindgrenj6:
After finding out that about `Exception#cause` (thanks fryguy!) I removed the `ManageIQ::API::Common::ApiError` class to move all the logic to a simple loop in the new `ExceptionHandling` mixin. 

// TODO: 
- [x] specs


lib/manageiq/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -0,0 +1,27 @@
...
+      module ApplicationControllerMixins
+        module ExceptionHandling
+          def self.included(other)
+            other.rescue_from(ManageIQ::API::Common::ApiError) do |exception|
+              errors = []
+              og_exception = exception
+
+              until exception.nil?
+                code = exception.respond_to?(:code) ? exception.code : exception.class.to_s
+                errors << {:status => code, :detail => exception.class.to_s + ": " + exception.message}
```
mkanoor: 👍 I like the exception class in the detail

lib/manageiq/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -0,0 +1,27 @@
+module ManageIQ
+  module API
+    module Common
+      module ApplicationControllerMixins
+        module ExceptionHandling
+          def self.included(other)
+            other.rescue_from(ManageIQ::API::Common::ApiError) do |exception|
```
mkanoor: @lindgrenj6 Why does this have to be ApiError, since most of the exceptions inherit from the StandardError should we be able to just use the StandardError.

lindgrenj6: I had thought about rescuing from standard error, but didn't want to be overly rescuing. Refactored a bit and I now have 2 rescue_from for StandardError and RuntimeError

lib/manageiq/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -0,0 +1,37 @@
...
+        module ExceptionHandling
+          def self.included(other)
+            other.rescue_from(StandardError, RuntimeError) do |exception|
+              errors, status = collected_ipp_errors(exception)
+
+              render :json => errors, :status => status
+            end
+          end
+
+          def collected_ipp_errors(exception, default_code = 400)
```
syncrou: I'm assuming `ipp_` stands for the Insights platform proposal? Maybe just `collected_errors` since that better defines the method behavior.

lib/manageiq/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -0,0 +1,37 @@
...
+            errors = []
+            og_exception = exception
```
syncrou: Not sure I understand the `og_` prefix either here.

lib/manageiq/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -0,0 +1,37 @@
...
+
+            until exception.nil?
+              code = exception.respond_to?(:code) ? exception.code : exception.class.to_s
+              errors << {:status => code, :detail => exception.class.to_s + ": " + exception.message}
+              exception = exception.cause
+            end
+
+            # Overwrite the first error since that one most likely doen't have a code
+            errors.first[:status] = if ActionDispatch::ExceptionWrapper.rescue_responses.key?(og_exception.class.to_s)
```
syncrou: Is there value moving this all into a method ( `assign_default_code` ) I wonder if it could return:
```
 [{:errors => errors}, errors.first[:status]]
```

lib/manageiq/api/common/application_controller_mixins/exception_handling.rb
lindgrenj6: Shorthand for original, changed it to `top_level_exception`

lindgrenj6:
@miq-bot add_reviewer bdunne

lindgrenj6:
@miq-bot assign bdunne

lindgrenj6:
@Fryguy @bdunne should I move this to a new PR? I know it changed a lot since I opened it - but it's ready to go now.

lib/manageiq/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -0,0 +1,43 @@
...
+            end
+          end
+
+          def collected_errors(exception)
+            errors = []
+            top_level_exception = exception
+
+            until exception.nil?
+              code = exception.respond_to?(:code) ? exception.code : exception.class.to_s
+              errors << {:status => code, :detail => exception.class.to_s + ": " + exception.message}
```
bdunne: Wouldn't this wind up setting `{:status => "SomeClass", ...}` if the exception doesn't respond to `#code`?

spec/dummy/config/routes.rb
```diff
@@ -9,6 +9,8 @@
     namespace :v1x0, :path => "v1.0" do
       get "/openapi.json", :to => "root#openapi"
       post "graphql" => "graphql#query"
+      get '/error', :to => "errors#error"
```
bdunne: Is this for test purposes rather than reusing an existing controller?

lib/manageiq/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -0,0 +1,43 @@
...
+              exception = exception.cause
+            end
+
+            # Overwrite the first error since that one most likely doen't have a code
+            errors.first[:status] = error_code_from_class(top_level_exception)
+
+            [{:errors => errors}, errors.first[:status]]
```
bdunne: The status for the document could be different from each of the errors in the body, right?

spec/dummy/config/routes.rb
lindgrenj6: Yes - I can move the routes to another controller if you would prefer that.

lib/manageiq/api/common/application_controller_mixins/exception_handling.rb
lindgrenj6: Yes - the status could be different but the goal is that we use the status that comes from the final Exception that was caught here (which we look up the proper code from on L27). It's kinda similar to how `ErrorDocument` works without having to pass that object around after the initial exception. 

One thing I could do is use the `error_code_from_class` method again if you would prefer that, it would be a bit more readable.

lib/manageiq/api/common/application_controller_mixins/exception_handling.rb
lindgrenj6: Yeah thats right, I'll use the error_code_from_class instead of just the exception name.

Fryguy: Prefer interpolation over `+` for string building

lib/manageiq/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -0,0 +1,43 @@
...
+
+            [{:errors => errors}, error_code_from_class(top_level_exception)]
+          end
+
+          def error_code_from_class(exception)
+            if ActionDispatch::ExceptionWrapper.rescue_responses.key?(exception.class.to_s)
+              ActionDispatch::ExceptionWrapper.rescue_responses[exception.class.to_s]
+            else
+              DEFAULT_ERROR_CODE
+            end
```
Fryguy: Can this be simplified to `ActionDispatch::ExceptionWrapper.rescue_responses[exception.class.to_s] || DEFAULT_ERROR_CODE` ?

lindgrenj6: See thats what I tried initially - but `ActionDispatch::ExceptionWrapper.rescue_responses` is a hash thats created with a default value of `:service_unavailable` so if the key doesn't exist it never falls back to the default code. 

src: https://github.com/rails/rails/blob/a3d54d2afe557b243d6b18f9adf119f60ffb8242/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb#L8

lindgrenj6:
@Fryguy @bdunne updated this to use the `ErrorDocument` class since I basically reimplemented it without thinking.

lindgrenj6:
@miq-bot remove_label unmergeable

bdunne:
@lindgrenj6 I added a couple of changes, can you review them?

PR #117, opened 2019-10-03
Add 404 "not found" to generated responses

gtanzillo:
Related to https://github.com/ManageIQ/sources-api/pull/95

https://projects.engineering.redhat.com/browse/TPINVTRY-232

### DIFF
--- a/lib/manageiq/api/common/open_api/generator.rb
+++ b/lib/manageiq/api/common/open_api/generator.rb
@@ -109,19 +109,19 @@ module ManageIQ
             klass_name = "ErrorNotFound"
 
             schemas[klass_name] = {
-              "type" => "object",
+              "type"       => "object",
               "properties" => {
-                "errors" => {
-                  "type" => "array",
+                "errors"  => {
+                  "type"  => "array",
                   "items" => {
-                    "type" => "object",
-                    "properties" => {
-                      "status" => {
-                        "type" => "integer",
+                    "type"        => "object",
+                    "properties"  => {
+                      "status"    => {
+                        "type"    => "integer",
                         "example" => 404
                       },
-                      "detail" => {
-                        "type" => "string",
+                      "detail"    => {
+                        "type"    => "string",
                         "example" => "Record not found"
                       }
                     }
@@ -213,9 +213,9 @@ module ManageIQ
 
               h["responses"]["404"] = {
                 "description" => "Not found",
-                "content" => {
+                "content"     => {
                   "application/json" => {
-                    "schema" => { "$ref" => build_schema_error_not_found }
+                    "schema"         => { "$ref" => build_schema_error_not_found }
                   }
                 }
               }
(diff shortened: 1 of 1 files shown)


lib/manageiq/api/common/open_api/generator.rb
```diff
@@ -176,6 +176,14 @@ def openapi_list_description(klass_name, primary_collection)
                       "schema" => { "$ref" => build_collection_schema(klass_name) }
                     }
                   }
+                },
+                "404" => {
+                  "description" => "#{klass_name.pluralize} collection",
+                  "content"     => {
+                    "application/json" => {
+                      "schema" => { "$ref" => build_collection_schema(klass_name) }
```
bdunne: This renders a 404 error document
```
{
  "errors": [
    {
      "status": 404,
      "detail": "Record not found"
    }
  ]
}
```

gtanzillo: @bdunne I made this change locally
```diff
diff --git a/lib/manageiq/api/common/open_api/generator.rb b/lib/manageiq/api/common/open_api/generator.rb
index 9efcbda..16528c7 100644
--- a/lib/manageiq/api/common/open_api/generator.rb
+++ b/lib/manageiq/api/common/open_api/generator.rb
@@ -176,16 +176,14 @@ module ManageIQ
                       "schema" => { "$ref" => build_collection_schema(klass_name) }
                     }
                   }
-                },
-                "404" => {
-                  "description" => "#{klass_name.pluralize} collection",
-                  "content"     => {
-                    "application/json" => {
-                      "schema" => { "$ref" => build_collection_schema(klass_name) }
-                    }
-                  },
                 }
-              }
+              },
+              "errors" => [
+                {
+                  "status": 404,
+                  "detail": "Record not found"
+                }
+              ]
```
And that build the 404 like this in the spec -
```
        },
        "errors": [
          {
            "status": 404,
            "detail": "Record not found"
          }
        ]
      },
```
But in other places int he spec we have it this way
```
          },
          "404": {
            "description": "Not found"
          }
        }
      },
```
Which one is correct?

bdunne: The first one is correct, we need the `"errors"` key at the top level of the response.

gtanzillo: Thanks, @bdunne, I'll work on making the necessary changes

lib/manageiq/api/common/open_api/generator.rb
```diff
@@ -176,6 +204,14 @@ def openapi_list_description(klass_name, primary_collection)
                       "schema" => { "$ref" => build_collection_schema(klass_name) }
                     }
                   }
+                },
+                "404" => {
```
bdunne: I think we want to skip this if `primary_collection` is `nil`, right?  We should only have a 404 if there is an id in the path.

PR #122, opened 2019-10-10
Added support for sort_by for sorting results.

abellotti:
Added support for sort_by for sorting results.

- Supports a single string representing an attribute, attribute:asc or attribute:desc
- Supports array of attribute, attribute:asc and attribute:desc

Examples:

URL Query:
- GET /api/v1.0/sources?sort_by=name
- GET /api/v1.0/vms?sort_by[]=power_state&sort_by[]=memory:desc

Ruby Client Parameter:
- :sort_by => "name"
- :sort_by => ["power_state", "memory:desc"]

GraphQL Parameter:
- sort_by: "name"
- sort_by: ["power_state, "memory:desc"]

GraphQL Example:

```
{
  sources(sort_by: ["source_type_id:asc", "uid:desc"]) {
    source_type_id
    id
    name
    uid
  }
}
```

Or as simple as:

```
{
  sources(sort_by: "name") {
    id
    name
  }
}
```

Enhancement as per TPINVTRY-705

### DIFF
--- a/lib/manageiq/api/common/application_controller_mixins/request_parameter_validation.rb
+++ b/lib/manageiq/api/common/application_controller_mixins/request_parameter_validation.rb
@@ -14,7 +14,7 @@ module ManageIQ
           def validate_request_parameters
             api_version = self.class.send(:api_version)[1..-1].sub(/x/, ".")
 
-            self.class.send(:api_doc).try(
+            api_doc.try(
               :validate_parameters!,
               request.method,
               request.path,
--- a/lib/manageiq/api/common/application_controller_mixins/request_parameter_validation.rb
+++ b/lib/manageiq/api/common/application_controller_mixins/request_parameter_validation.rb
@@ -1,0 +1,29 @@
+module ManageIQ
+  module API
+    module Common
+      module ApplicationControllerMixins
+        module RequestParameterValidation
+          def self.included(other)
+            other.include(OpenapiEnabled)
+
+            other.before_action(:validate_request_parameters)
+          end
+
+          private
+
+          def validate_request_parameters
+            api_version = self.class.send(:api_version)[1..-1].sub(/x/, ".")
+
+            self.class.send(:api_doc).try(
+              :validate_parameters!,
+              request.method,
+              request.path,
+              api_version,
+              params.slice(:sort_by)
+            )
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/requests/api/v1.0/filter_spec.rb
+++ b/spec/requests/api/v1.0/filter_spec.rb
@@ -71,12 +71,12 @@ RSpec.describe("::ManageIQ::API::Common::Filter", :type => :request) do
   end
 
   context "sorted results via sort_by" do
-    let!(:rhev)      { SourceType.create(:name => "rhev", :product_name => "RedHat Virtualization", :vendor => "redhat") }
-    let!(:openstack) { SourceType.create(:name => "openstack", :product_name => "OpenStack", :vendor => "redhat") }
-    let!(:vmware)    { SourceType.create(:name => "vmware", :product_name => "OpenStack", :vendor => "vmware") }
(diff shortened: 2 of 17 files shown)


abellotti:
[WIP] needs the following:

 - [ ] OpenAPI updates
 - [ ] REST API specs
 - [ ] GraphQL specs

gmcculloug:
@Hyperkid123 Another PR to keep an eye on.  Currently catalog is enforcing ordering of the `Order`instance in the model and I would like to remove that and allow the UI to define the ordering.  (Same for `Request` on Approval service)

https://github.com/ManageIQ/catalog-api/blob/master/app/models/order.rb#L8
```ruby
default_scope { kept.order(:created_at => :desc) }
```

Fryguy:
@abellotti Is this based on some standard or convention or is this made up?  I'd rather follow some convention if possible, that way if we have to switch / bring to the platform, it's not a huge change if any.

abellotti:
@Fryguy No standards out there for REST, but many implementations.  When I was investigating there there seems to be about 4 conventions, they are:

- attribute, +attribute and -attribute (+ being the asc case, and - the desc)
- attribute, attribute:asc and attribute:desc
- attribute, attribute.asc, attribute.desc
- separate attribute and order pairs  (sort and order), kinda's clunky

References: https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/#

I tended to lean toward the second convention, it is a convention used by OpenStack Reference: https://specs.openstack.org/openstack/api-wg/guidelines/pagination_filter_sort.html#sorting 

Although I have the sort_by[] for the list, I can switch to the single comma separated string, then it would be an exact implementation if you prefer.

sort_by="attribute[:asc|:desc][,attribute[:asc|:desc][,...]"  

it would also make it easier on the different clients.

thoughts ?

abellotti:
ping @Fryguy, other than an update to the README, this PR is good to unwip. thoughts on chosen convention ?

bdunne:
> Although I have the sort_by[] for the list, I can switch to the single comma separated string,

I vote for staying with `sort_by[]=attribute&sort_by[]=attribute2:desc` to be consistent with the filtering API

spec/requests/api/v1.0/filter_spec.rb
```diff
@@ -18,6 +18,29 @@ def expect_success(query, *results)
...
+    )
+  end
+
+  def expect_success_ordered_objects(collection, query, results)
+    get(URI.escape("/api/v1.0/#{collection}?#{query}"))
+
+    expect(response.status).to(eq(200))
+
+    response.parsed_body["data"].each do |object|
+      expect(object).to(include(results.shift.deep_stringify_keys))
```
Fryguy: Something about this helper feels weird to me.  I would expect it to return an order set that the rest itself can compare with a normal `eq` matcher

lib/manageiq/api/common/application_controller_mixins/request_parameter_validation.rb
```diff
@@ -0,0 +1,29 @@
...
+
+            other.before_action(:validate_request_parameters)
+          end
+
+          private
+
+          def validate_request_parameters
+            api_version = self.class.send(:api_version)[1..-1].sub(/x/, ".")
+
+            self.class.send(:api_doc).try(
```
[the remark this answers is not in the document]
abellotti: it's a private method of self.class, alternatively I think I can just call api_doc.try(....)

spec/requests/api/v1.0/filter_spec.rb
abellotti: I had to do something like this here because eq matches the whole hash, but I only cared to match a subset.  For using eq for this purpose, I guess one could do the following:

```
 results = results.collect(&:stringify_keys)
 attrs = results.first.keys

 expect(response.parsed_body["data"].collect { |res| res.slice(*attrs) }).to(eq(results))
```

abellotti: This also allows us to do a single expect vs. one per resource returned.

PR #123, opened 2019-10-10
Adding support for case insensitive searches.

abellotti:
Adding support for case insensitive searches.

Approach taken here to support the case insensitive operators for the filter parameter in addition to the current string operators.

So adding support for:

  - starts_with_i
  - ends_with_i
  - contains_i
  - eq_i

This works with both REST API as well as GraphQL

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,26 @@ Or install it yourself as:
 
 #### ManageIQ::Api::Common::Filter
 
+| Supported Comparators     | Comparator    |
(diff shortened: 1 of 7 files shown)


abellotti:
/cc @gmcculloug ask and ye-shall receive.

WIP for now, needing README and specs updates.

gmcculloug:
Thanks @abellotti 

cc @Hyperkid123

PR #125, opened 2019-10-17
Fix error responses for filtering

bdunne:
After upgrading to manageiq-api-common v2.0.0 I got error details like:
"ManageIQ::API::Common::Filter::Error: ManageIQ::API::Common::Filter::Error"

With this change they're more like we would expect:
"ManageIQ::API::Common::Filter::Error: found unpermitted parameter: xxx"

### DIFF
--- a/lib/manageiq/api/common/filter.rb
+++ b/lib/manageiq/api/common/filter.rb
@@ -23,11 +23,11 @@ module ManageIQ
               type = determine_string_attribute_type(attribute)
               send(type, k, v)
             else
-              errors.add("unsupported attribute type for: #{k}")
+              errors << "unsupported attribute type for: #{k}"
             end
           end
 
-          raise ManageIQ::API::Common::Filter::Error.new(:error_document => errors) unless errors.blank?
+          raise(ManageIQ::API::Common::Filter::Error, errors.join(", ")) unless errors.blank?
           query
         end
 
@@ -35,18 +35,12 @@ module ManageIQ
 
         attr_accessor :query
 
-        class Error < ArgumentError
-          attr_reader :error_document
-
-          def initialize(attrs)
-            @error_document = attrs[:error_document]
-          end
-        end
+        class Error < ArgumentError; end
 
         def attribute_for_key(key)
           attribute = api_doc_definition.properties[key.to_s]
           return attribute if attribute
-          errors.add("found unpermitted parameter: #{key}")
+          errors << "found unpermitted parameter: #{key}"
           nil
         end
 
@@ -57,7 +51,7 @@ module ManageIQ
         end
 
         def errors
-          @errors ||= ManageIQ::API::Common::ErrorDocument.new
+          @errors ||= []
         end
 
         def string(k, val)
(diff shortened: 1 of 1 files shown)


lib/manageiq/api/common/filter.rb
```diff
@@ -23,30 +23,24 @@ def apply
               type = determine_string_attribute_type(attribute)
               send(type, k, v)
             else
-              errors.add("unsupported attribute type for: #{k}")
+              errors << "unsupported attribute type for: #{k}"
             end
           end
 
-          raise ManageIQ::API::Common::Filter::Error.new(:error_document => errors) unless errors.blank?
+          raise(ManageIQ::API::Common::Filter::Error, errors.join(", ")) unless errors.blank?
```
bdunne: I'm joining the errors with `", "` now unless you see some other way to add multiple errors @lindgrenj6 

Here's my diff in topo after upgrading:
```diff
       expect(response.parsed_body["errors"]).to eq(
         [
-          {"detail" => "unsupported attribute type for: mac_addresses", "status" => 400},
-          {"detail" => "found unpermitted parameter: bogus_attribute", "status" => 400}
+          {"detail" => "ManageIQ::API::Common::Filter::Error: unsupported attribute type for: mac_addresses, found unpermitted parameter: bogus_attribute", "status" => 400}
         ]
       )
```

mkanoor:
@bdunne @abellotti  We were following the IPP for error reporting which is based on the JSON-API spec. And errors is supposed to be an array of error messages
https://jsonapi.org/examples/#error-objects-multiple-errors

bdunne:
Yes, I originally wrote it to support returning an array of errors by adding errors to the `ErrorDocument`.  

With the change in error handling, the new API is to `raise` you error and that will be caught in a single place and an `Error Document` will be created from that.  (BTW, that is so much cleaner than having `rescue_from` all over the place)

So, did we lose the ability to add multiple messages to an error? Or maybe I am misunderstanding something? @lindgrenj6

mkanoor:
@bdunne This spec shows us handling an array of error messages
https://github.com/ManageIQ/manageiq-api-common/blob/9d9fd09cc376bb50d9e1bfc2cffe20ea005468a8/spec/requests/exception_handling_spec.rb#L25

lib/manageiq/api/common/filter.rb
```diff
@@ -23,30 +23,24 @@ def apply
               type = determine_string_attribute_type(attribute)
               send(type, k, v)
             else
-              errors.add("unsupported attribute type for: #{k}")
+              errors << "unsupported attribute type for: #{k}"
```
mkanoor: @bdunne This pattern seems to be like how ActiveRecord does the validations of all parameters and collects them and raises a single error if the @errors array is not empty. With the other exception handling we are raising a chain of exceptions and traversing the exception cause to build an array of error messages.
https://github.com/rails/rails/blob/6d68bb5f695414b1801c1c3952ef955a5b0b6c6a/activerecord/lib/active_record/validations.rb#L21

PR #126, opened 2019-10-18
We missed the changes needed for seeding

mkanoor:
This PR https://github.com/ManageIQ/catalog-api/pull/391 was
created after we had the PR to move the RBAC into the common
gem


mkanoor:
@bdunne Can you review merge this PR it will break the builds for Catalog if we switch to the newer 2.0 gem

PR #127, opened 2019-10-22
[WIP] Use indexed arrays in request for filtering

lpichler:
there is issue in handling indexed arrays in requests:

`/api/vms/filter[id][0]=2&filter[id][1]=44`
`/api/vms/filter[id][eq][0]=2&filter[id][eq][1]=44`
`...`

because in rails servers , there is  in `params`:  ( array is transformed to the hash as expected):
`{'0' => '1', '1' => '44'}`

I know that according to documentation we are not using this form but this exactly uses `Typhoeus   (EasyFactory) gem` and `Typhoeus gem ` is used in **our generated ruby Topological/Sources clients.**

Fix convert **such hashes** (`{'0' => '1', '1' => '44'}`) to arrays ['1', '44'] when **such hashes**  are present in params.

#### Testing scripts with for ruby Topological and RestClient(uses query: `filter[id][][]=1&filter[id][]=a`)

```ruby
TopologicalInventoryApiClient.configure { |config| config.scheme = "http"; config.host = "localhost:3000" }
 


# TopologicalInventoryApiClient

api_client = TopologicalInventoryApiClient::ApiClient.new
tap = TopologicalInventoryApiClient::DefaultApi.new(api_client)
ih = {'x-rh-identity' => Base64.strict_encode64(JSON.dump('identity' => { 'account_number' => 'XXX' }))}
api_client.default_headers.merge!(ih)

tap.list_vms({:filter => {:id =>[1,2,3]}}) 
# ^^^ This was throwing an error(without this fix) and it produces query:
# /api/vms/filter[id][0]=1&filter[id][1]=2&filter[id][1]=3


# RESTCLIENT
require 'rest-client'

tph = {:host => 'localhost', :port => 3000, :path => '/api/topological-inventory/v1.0/vms',  :query => {:filter => { :id => [1,2,3]} }.to_query}
  
uri = URI::HTTP.build(tph).to_s

ih = {'x-rh-identity' => Base64.strict_encode64(JSON.dump('identity' => { 'account_number' => 'XXX' }))}
   
RestClient::Request.execute(:method => :get, :url => uri.to_s, :headers => ih)
# this produces query which we can handle:
/api/vms/filter[id][]=1&filter[id][]=2&filter[id][]=3
```

so this fix allows us to use generated clients for batch query.

what do you think about this fix ? @abellotti @bdunne @slemrmartin 

thanks!

### DIFF
--- a/lib/manageiq/api/common/filter.rb
+++ b/lib/manageiq/api/common/filter.rb
@@ -14,6 +14,34 @@ module ManageIQ
           @api_doc_definition = api_doc_definition
         end
 
+        def array_of_integers?(values)
+          return false if values.blank?
+
+          values.all? do |v|
+            begin
+              Integer(v)
+            rescue StandardError
+              nil
+            end.present?
+          end
+        end
+
+        def handle_arrays_in_params_with_comparators(hash_params)
+          hash_params.to_h.each_with_object(hash_params).each do |(key, value), return_hash|
+            return_hash[key] = array_of_integers?(value.try(:keys)) ? value.values : value
+          end
+        end
+
+        def normalize_arrays_in_params(hash_params)
+          indexes_or_comparators = hash_params.try(:keys)
+          return hash_params unless indexes_or_comparators
+
+          are_indexes = array_of_integers?(indexes_or_comparators)
+          return hash_params.values if are_indexes
+
+          handle_arrays_in_params_with_comparators(hash_params)
+        end
+
         def apply
           return query if @raw_filter.blank?
           @raw_filter.each do |k, v|
@@ -21,6 +49,7 @@ module ManageIQ
 
             if attribute["type"] == "string"
               type = determine_string_attribute_type(attribute)
+              v = normalize_arrays_in_params(v)
               send(type, k, v)
             else
               errors.add("unsupported attribute type for: #{k}")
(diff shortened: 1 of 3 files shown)


bdunne:
I'd rather fix the client so that it sends an array rather than a hash (indexed array).  I'm afraid that we'll want to send a hash in the future and this could get in the way.

Is this a common standard?  I know that there's no "official" specification for arrays in the query string, but it wasn't one of the 3 popular examples that I found (below).
```ruby
?id[]=1&id[]=2&id[]=3
=> {'id' => [1,2,3]}

?id=1&id=2&id=3
=> {'id' => 3}

?id=1,2,3
=> {'id' => "1,2,3"}
```

lpichler:
@bdunne thank you for your input.

so I was continuing with more deeper investigation and we can use clients with this setting and then it will work for arrays). (example: `client.list_vms({:filter => {:id =>[1,2,3]}})`)

```ruby
TopologicalInventoryApiClient.configure do |config|
   config.scheme = “http”; config.host = "localhost:3000"
   config.params_encoding = :rack # <---- setting 
end
``` 

maybe can we set it _somehow_ as default for `config.params_encoding ` (not sure what else can this setting affect ) ? 

NOTE: `params_encoding` refers  to this [line](https://github.com/typhoeus/ethon/blob/master/lib/ethon/easy/queryable.rb#L101). (previously there was a `nil`)

I will change this PR to just add some few specs.

lpichler:
specs moved here https://github.com/ManageIQ/manageiq-api-common/pull/128

PR #129, opened 2019-10-23
Fix nested attributes

agrare:
The OpenApi::Docs::ObjectDefinition.all_attributes method was only
returning top level attribute names without considering that some of
those attributes might be nested objects.

This was causing the body_params.permit! to fail because it found nested
parameters.

https://github.com/ManageIQ/sources-api/issues/110


agrare:
Thanks @bdunne , can I get a release for https://github.com/ManageIQ/sources-api/pull/112 ?

bdunne:
v2.0.1 has been released

PR #130, opened 2019-10-24
Changed scope for sharing/unsharing to be account based

mkanoor:
When the RBAC code was created in common gem it  was from the Approval source code
which uses the default scope to be principal (currently logged in user).
For Catalogs Sharing/Unsharing/Query feature we need to be in tenant
scope. This PR passes in the proper scope


mkanoor:
@bdunne Please review

bdunne:
@hsong-rh Please review

PR #132, opened 2019-11-06
Rename gem to insights-api-common

Fryguy:
@bdunne @agrare @gtanzillo @gmcculloug @syncrou Please review.

😱


Fryguy:
All tests passed...not sure if that means 100% coverage though

syncrou:
@Fryguy do we want all of the change log to say insights_ vs manageiq_ for all the entries including those in the past? Possibly track the name change in the change log?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Fryguy:
This PR has the name change on the changelog but I didn't go back and change old entries.  Not sure if I should...I debated it.  Curious what others think

agrare:
I don't think we should go back and change old CHANGELOG entries, :+1: for what you have here

CHANGELOG.md
```diff
@@ -56,11 +58,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-[1.0.1]: https://github.com/ManageIQ/manageiq-api-common/compare/v1.0.0...v1.0.1
-[1.0.0]: https://github.com/ManageIQ/manageiq-api-common/compare/v0.1.0...v1.0.0
-[0.1.0]: https://github.com/ManageIQ/manageiq-api-common/releases/tag/v0.1.0
+[Unreleased]: https://github.com/RedHatInsights/insights-api-common-rails/compare/v2.0.1...HEAD
+[2.0.1]: https://github.com/RedHatInsights/insights-api-common-rails/compare/v2.0.0...v2.0.1
+[2.0.0]: https://github.com/RedHatInsights/insights-api-common-rails/compare/v1.1.0...v2.0.0
+[1.1.0]: https://github.com/RedHatInsights/insights-api-common-rails/compare/v1.0.2...v1.1.0
+[1.0.2]: https://github.com/RedHatInsights/insights-api-common-rails/compare/v1.0.1...v1.0.2
+[1.0.1]: https://github.com/RedHatInsights/insights-api-common-rails/compare/v1.0.0...v1.0.1
+[1.0.0]: https://github.com/RedHatInsights/insights-api-common-rails/compare/v0.1.0...v1.0.0
```
syncrou: In regards to the CHANGELOG comments earlier, possibly I'm misunderstanding the `[Unreleased]` heading.  I would have expected line 67 to still say `ManageIQ` since on line 39 it was released before this name change PR. However if the `[Unreleased]` heading needs the current 'real' name/location of the repo then these changes make sense.

Fryguy: These lines at the bottom are links and when we move the repo they will no longer be valid.  I have to change these.  What I was referring to were references to ManageIQ::Api constants elsewhere in the content, and those I left alone.

Fryguy:
@bdunne After this is merged, I'd like to release this as a major version bump.  Then, the apps can update accordingly.  We can move the repo anytime after merge since if teams are using the gem, the code location really doesn't matter.

gmcculloug:
Is `Insights` the proper name to switch our modules and other items to?

I realize the repo name we are moving to is [RedHatInsights](https://github.com/RedHatInsights) but that name was selected when the platform was referred to as Insights.  Now the platform is commonly referred to as `Cloud Services` or `Cloud Platform` or `cloud.redhat.com` and an application uses the `Insights` name.

I'm fine if we want to keep `Insights` but thought it was worth pointing out.

Fryguy:
It's a good question, and that did occur to me as well, but I can't think of what else to call it.  As you said, insights is the org, so that seemed ok to me.

Fryguy:
What do you all think of `RedHatCloudServices::Api::Common` & `red_hat_cloud_services-api-common` ?

@bdunne @gtanzillo @syncrou @gmcculloug @agrare ?

agrare:
That works too, most of the platform repos there are still called `insights-` though so I don't think its a big deal either way.

Fryguy:
> most of the platform repos there are still called insights-

that was actually one of the other reasons I went with insights, but I think insights is still a "product" that is a subset of cloud.redhat.com

gtanzillo:
The thing is that it will probably change multiple times over the life of this repo so either one is fine.

Actually, "insights" is probably safer because it is an existing product, as you mentioned, and there are existing repos with that name.

agrare:
> but I think insights is still a "product" that is a subset of cloud.redhat.com

Oh yeah not disagreeing with that at all, just noting that most of the repos from "that platform formerly known as insights" are still called "insights-" even if they aren't for "inisghts, the app"

Fryguy:
I'm leaning towards keeping it `insights-` personally, aside from the naming, it's just shorter.

agrare:
:+1:

bdunne:
I'm also fine with "insights-"

bdunne:
@Fryguy Sorry for the merge conflict.  I had to release a new version before we renamed.

CHANGELOG.md
bdunne: Should the change to `RedHatInsights` happen after the transfer?

lib/insights/api/common/application_controller_mixins/api_doc.rb
```diff
@@ -21,7 +21,7 @@ module ClassMethods
             private
 
             def api_doc
-              @api_doc ||= ::ManageIQ::API::Common::OpenApi::Docs.instance[api_version[1..-1].sub(/x/, ".")]
```
bdunne: There was a namespace issue here, that's why this is fully qualified

lib/insights/api/common/graphql/generator.rb
```diff
@@ -79,15 +79,15 @@ def self.collection_schema_overlay(schema_overlay, collection)
...
 
           def self.init_schema(request, schema_overlay = {})
-            api_version       = ::ManageIQ::API::Common::GraphQL.version(request)
+            api_version       = GraphQL.version(request)
             version_namespace = "V#{api_version.tr('.', 'x')}"
-            openapi_doc       = ::ManageIQ::API::Common::OpenApi::Docs.instance[api_version]
+            openapi_doc       = OpenApi::Docs.instance[api_version]
             openapi_content   = openapi_doc.content
 
-            graphql_namespace = if ::ManageIQ::API::Common::GraphQL::Api.const_defined?(version_namespace, false)
```
bdunne: I think there was also a namespace issue here

lib/insights/api/common/application_controller_mixins/api_doc.rb
Fryguy: I'm going to back out all of the namespace changes...can always do that separately

CHANGELOG.md
Fryguy: I think it can go ahead of time...it's just docs and it's ok if they temporarily don't resolve correctly.

Fryguy:
@bdunne I removed all of the namespace manipulations and put them back, so this is now just a pure rename.

README.md
```diff
@@ -1,20 +1,18 @@
...
-[![Maintainability](https://api.codeclimate.com/v1/badges/790ea6c77d82da6be68a/maintainability)](https://codeclimate.com/github/ManageIQ/manageiq-api-common/maintainability)
-[![Test Coverage](https://api.codeclimate.com/v1/badges/790ea6c77d82da6be68a/test_coverage)](https://codeclimate.com/github/ManageIQ/manageiq-api-common/test_coverage)
-[![Security](https://hakiri.io/github/ManageIQ/manageiq-api-common/master.svg)](https://hakiri.io/github/ManageIQ/manageiq-api-common/master)
+[![Build Status](https://travis-ci.org/RedHatInsights/insights-api-common-rails.svg)](https://travis-ci.org/RedHatInsights/insights-api-common-rails)
+[![Maintainability](https://api.codeclimate.com/v1/badges/790ea6c77d82da6be68a/maintainability)](https://codeclimate.com/github/RedHatInsights/insights-api-common-rails/maintainability)
+[![Test Coverage](https://api.codeclimate.com/v1/badges/790ea6c77d82da6be68a/test_coverage)](https://codeclimate.com/github/RedHatInsights/insights-api-common-rails/test_coverage)
+[![Security](https://hakiri.io/github/RedHatInsights/insights-api-common-rails/master.svg)](https://hakiri.io/github/RedHatInsights/insights-api-common-rails/master)
```
lpichler: https://hakiri.io/github/RedHatInsights/insights-api-common-rails/master 
this link is not working and there is similar issue on some topo repos. If I can make it work let me know or I will create issue/jira issue to track this issue somewhere.

gmcculloug: @lpichler I believe either myself or @gtanzillo can update the hakiri webhook on the repos.  I suggest opening an issue so we can work through updating the other repos.

lpichler: it looks that it is fixed for this repo, thanks 👍 

anyway I created separated issues for other repos where it needs to be fixed:
https://github.com/RedHatInsights/topological_inventory-api-client-ruby/issues/27
https://github.com/RedHatInsights/topological_inventory-ansible_tower/issues/68
https://github.com/RedHatInsights/topological_inventory-amazon/issues/57

PR #133, opened 2019-11-08
Fix params for update to support Hash/Array

hsong-rh:


hsong-rh:
@bdunne Please review and merge

lib/insights/api/common/application_controller_mixins/parameters.rb
```diff
@@ -104,7 +104,23 @@ def query_sort_by
 
           def params_for_update
             check_if_openapi_enabled
-            body_params.permit(*api_doc_definition.all_attributes - api_doc_definition.read_only_attributes)
+            attr_list = *api_doc_definition.all_attributes - api_doc_definition.read_only_attributes
+            strong_params_hash = hashes_and_arrays(api_doc_definition, attr_list)
+            body_params.permit(strong_params_hash)
+          end
+
+          def hashes_and_arrays(api_doc_definition, attributes)
```
mkanoor: @hsong-rh I think we might want to come up with a better name for this function.
I am thinking **sanctified_permit_param** since it has regular attributes as well as hashes and arrays

mkanoor: We should also add a comment that this been already validated by the openapi parser and this is just to satisfy the strong parameters check

hsong-rh: Changed!

PR #134, opened 2019-11-13
Allow for other content types other than application/json

### LINKED ISSUE #116: JSON Parser error when trying to parse multipart request body
We currently assume that the body is always json without first checking the content type and fail with a JSON parser error.

https://github.com/ManageIQ/manageiq-api-common/blob/382522526586070fa7c07f9bbad04854e6f84864/lib/manageiq/api/common/application_controller_mixins/request_body_validation.rb#L35

Openapi allows for Files to be uploaded using the multipart/form-data content type

https://swagger.io/docs/specification/describing-request-body/multipart-requests/

mkanoor:
Rails can handle multiple content types like application/json,
multipart/form-data. It parses them and builds the ActionController::
Parameters with the params that includes the path_parameters and
query_parameters. During a post call where we try to get just the
body parameters by reparsing the body, if we just exclude the path
parameters from the params we wont have to reparse the request.body.
Also the correct content-type needs to be sent to the openapi-parser.

Fixes #116


spec/dummy/config/initializers/wrap_parameters.rb
```diff
@@ -1,14 +0,0 @@
...
-
-# Enable parameter wrapping for JSON. You can disable this by setting :format to an empty array.
-ActiveSupport.on_load(:action_controller) do
-  wrap_parameters format: [:json]
-end
-
-# To enable root element in JSON for ActiveRecord objects.
-# ActiveSupport.on_load(:active_record) do
-#   self.include_root_in_json = true
-# end
```
mkanoor: This file had to be removed because it tries to wrap the JSON parameters into an object based on the controller name. In our case we are expecting the openapi-parser to do the validation for us and it doesn't like the wrapped object which is an extra parameter and it rejects it. None of our other apps catalog, approval, sources and topology are using the wrap parameters feature of rails.

lib/insights/api/common/application_controller_mixins/request_body_validation.rb
```diff
@@ -18,11 +18,9 @@ def self.included(other)
...
           def body_params
             @body_params ||= begin
-              raw_body    = request.body.read
-              parsed_body = raw_body.blank? ? {} : JSON.parse(raw_body)
-              ActionController::Parameters.new(parsed_body).permit!
-            rescue JSON::ParserError
-              raise Insights::API::Common::ApplicationControllerMixins::RequestBodyValidation::BodyParseError, "Failed to parse request body, expected JSON"
+              hash = params.permit!.to_h
+              request.path_parameters.keys.each { |key| hash.delete(key) }
+              ActionController::Parameters.new(hash).permit!
```
bdunne: What do you think of this?
```
ActionController::Parameters.new(
	params.permit!.to_h.except(request.path_parameters.keys)
).permit!
```

lib/insights/api/common/application_controller_mixins/request_body_validation.rb
```diff
@@ -38,7 +36,8 @@ def validate_request
               request.method,
               request.path,
               api_version,
-              body_params.as_json
+              body_params.to_h,
```
bdunne: I don't remember the specifics, but I think there is a difference between `as_json` and `to_h`.

lib/insights/api/common/application_controller_mixins/request_body_validation.rb
```diff
@@ -18,11 +18,9 @@ def self.included(other)
 
           def body_params
             @body_params ||= begin
-              raw_body    = request.body.read
-              parsed_body = raw_body.blank? ? {} : JSON.parse(raw_body)
-              ActionController::Parameters.new(parsed_body).permit!
-            rescue JSON::ParserError
-              raise Insights::API::Common::ApplicationControllerMixins::RequestBodyValidation::BodyParseError, "Failed to parse request body, expected JSON"
+              hash = params.permit!.to_h
```
bdunne: What other content-types do you want to allow?  What happens when you call `.to_h` on them?

mkanoor: @bdunne 
For catalog we need to support application/json and multipart/form-data as listed here in the openapi

https://swagger.io/docs/specification/describing-request-body/multipart-requests/

lib/insights/api/common/application_controller_mixins/request_body_validation.rb
mkanoor: @bdunne 
The to_h returns a Hash similar to as_json but it changes the attribute values

For multipart/form-data we have a UploadedFile object that gets lost when we do the as_json

(byebug) j.to_h
{"content"=>#<ActionDispatch::Http::UploadedFile:0x00007fed85a2b438 @tempfile=#<Tempfile:/var/folders/kr/3m9v5qk557970qwndrlkdxyr0000gn/T/RackMultipart20191114-11952-19uzl2c.svg>, @original_filename="ocp_logo.svg", @content_type="image/svg+xml", @headers="Content-Disposition: form-data; name=\"content\"; filename=\"ocp_logo.svg\"\r\nContent-Type: image/svg+xml\r\nContent-Length: 21247\r\n">, "source_id"=>"27", "source_ref"=>"icon_ref", "portfolio_item_id"=>"2755"}

(byebug) j.as_json
{"content"=>{"tempfile"=>"#<File:0x00007fed7f53f1a0>", "original_filename"=>"ocp_logo.svg", "content_type"=>"image/svg+xml", "headers"=>"Content-Disposition: form-data; name=\"content\"; filename=\"ocp_logo.svg\"\r\nContent-Type: image/svg+xml\r\nContent-Length: 21247\r\n"}, "source_id"=>"27", "source_ref"=>"icon_ref", "portfolio_item_id"=>"2755"}

PR #135, opened 2019-11-13
Add check for readOnly parameters in `params_for_create`

lindgrenj6:
https://projects.engineering.redhat.com/browse/SSP-927

Found a bug when investigating this, basically in the `params_for_create` method before we were just allowing all the fields to come through whether they were readOnly or not. `params_for_update` did the right thing, but `params_for_create` was a bit more complicated since the requstBody can come in as a different ref with different fields.

This PR enhances the `params_for_create` function in the ParameterMixin so that it checks for any readOnly parameters as well as the requestBody schema if passing in `:writeable => true` when calling `params_for_create`.

### DIFF
--- a/lib/insights/api/common/application_controller_mixins/parameters.rb
+++ b/lib/insights/api/common/application_controller_mixins/parameters.rb
@@ -7,10 +7,25 @@ module Insights
             other.include(OpenapiEnabled)
           end
 
-          def params_for_create
+          def params_for_create(opts = {})
             check_if_openapi_enabled
             # We already validate this with OpenAPI validator, that validates every request, so we shouldn't do it again here.
-            body_params.permit!
+            if opts[:writeable] == true
+              attr_list = api_doc_definition.all_attributes - api_doc_definition.read_only_attributes
+              body_params.permit(*request_body_schema_keys + attr_list)
+            else
+              body_params.permit!
+            end
+          end
+
+          def request_body_schema_keys
+            Insights::API::Common::OpenApi::Docs.instance[api_version].content.dig("components", "schemas", request_body_schema_name, "properties").keys
+          rescue StandardError
+            []
+          end
+
+          def request_body_schema_name
+            Insights::API::Common::OpenApi::Docs.instance[api_version].content.dig("paths", "/#{controller_name}", "post", "requestBody", "content", "application/json", "schema", "$ref").split("/").last
           end
 
           def safe_params_for_list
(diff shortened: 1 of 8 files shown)


lindgrenj6:
cc @syncrou @mkanoor

lindgrenj6:
Yeah this probably won't work for those services, it's going to need to be a separate method. `writeOnce` would really fix everything here - but the closest thing to the spec is `writeOnly` which is definitely not what we want!

After discussing with @syncrou and @mkanoor, I'm going to move the logic I have here into a separate function or add a parameter to `params_for_create` to return write-only params so we can use that in catalog.

lindgrenj6:
@bdunne updated so it won't break sources/topology. This way anyone who wants to have POST requests validate against any read-only parameters this will do it.

lib/insights/api/common/application_controller_mixins/parameters.rb
```diff
@@ -7,10 +7,25 @@ def self.included(other)
...
+            if opts[:writeable] == true
+              attr_list = api_doc_definition.all_attributes - api_doc_definition.read_only_attributes
+              body_params.permit(*request_body_schema_keys + attr_list)
+            else
+              body_params.permit!
+            end
+          end
+
+          def request_body_schema_keys
+            Insights::API::Common::OpenApi::Docs.instance[api_version].content.dig("components", "schemas", request_body_schema_name, "properties").keys
```
mkanoor: @lindgrenj6 Is the because the required properties are on the body instead of in the object definitions. What would happen if it's defined in both the object and in body.
https://github.com/RedHatInsights/catalog-api/blob/10f2396699672a844c4064c089bd3b4e7f28b3d9/public/doc/openapi-3-v1.0.0.json#L129

lindgrenj6: Nothing would change - this is looking up the schema reference found in the `requestBody` parameter. 

It takes the reference here:
https://github.com/RedHatInsights/catalog-api/blob/10f2396699672a844c4064c089bd3b4e7f28b3d9/public/doc/openapi-3-v1.0.0.json#L856-L865

And looks up the schema here:
https://github.com/RedHatInsights/catalog-api/blob/10f2396699672a844c4064c089bd3b4e7f28b3d9/public/doc/openapi-3-v1.0.0.json#L2843-L2853

lindgrenj6: The openapi parser does the validation on the required keys - that isn't changing. This just checks the readOnly keys as well as looks up any extra ones specified in a `requestBody` schema.

PR #137, opened 2019-11-25
[WIP] Removed references to Policy objects

mkanoor:
https://projects.engineering.redhat.com/browse/SSP-963
Depends on https://github.com/RedHatInsights/insights-rbac-api-client-ruby/pull/4
Based on https://github.com/RedHatInsights/insights-rbac/pull/156

Instead of adding roles to group via Policies now we can directly add
role to a group or delete a role from a group


Gemfile
```diff
@@ -6,6 +6,5 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 # Declare your gem's dependencies in topological_inventory.gemspec.
 # Bundler will treat runtime dependencies like base dependencies, and
 # development dependencies will be added by default to the :development group.
-gem 'rbac-api-client', :git => "https://github.com/RedHatInsights/insights-rbac-api-client-ruby", :branch => "master"
-
+gem 'rbac-api-client', :git => "https://github.com/mkanoor/insights-rbac-api-client-ruby", :branch => "remove_policy"
```
mkanoor: I had to do this because I am waiting for this PR to get merged.
https://github.com/RedHatInsights/insights-rbac-api-client-ruby/pull/4

spec/support/rbac_seed_context.rb
```diff
@@ -0,0 +1,19 @@
+RSpec.shared_context "rbac_seed_objects" do
+  let(:app_name) { 'catalog' }
+  let(:resource) { "portfolios" }
+  let(:permissions) { ["#{app_name}:#{resource}:read"] }
```
eclarizio: I don't think this is being used at all? I think it can go away.

spec/support/rbac_seed_context.rb
```diff
@@ -0,0 +1,19 @@
...
+  let(:resource_id1) { "99" }
```
eclarizio: This could be removed and replaced inline on line 13 I think, there's nothing checking this value so seems superfluous to promote it to a `let`

mkanoor:
@eclarizio @syncrou 
Please review

mkanoor:
@gmcculloug @syncrou 
Since this PR removes Policy objects we could delay this PR and create a new PR that uses add_roles_to_group only during seeding. Since we would be using local AccessControlEntries (ACE) we don't need to be adding the sharing roles into the group anymore. This will allow us to stay backward compatibility and have the new sharing implemented locally. If we agree I can pull out the pieces from this PR that uses the new add_roles_to_group for the seeding process.

mkanoor:
Closed in favor of https://github.com/RedHatInsights/insights-api-common-rails/pull/138 which has the seeding pieces

PR #143, opened 2020-01-07
Add generic StatusController that can be mapped to `/health` for liveness probe across services

lindgrenj6:
This work was initially implemented and reviewed here: https://github.com/RedHatInsights/catalog-api/pull/528

This is basically the extraction of the (albeit tiny) controller that can then be added to any application and then mapped to `/health` as in the dummy app here. The goal is to use this across the microservices to have a liveness probe watch this URL.


lindgrenj6:
cc @carbonin

carbonin:
@bdunne want to take a look?

lib/insights/api/common/status.rb
```diff
@@ -0,0 +1,50 @@
...
+          end
+
+          private
+
+          def migration_context
+            ActiveRecord::Base.connection.migration_context
+          end
+
+          def pending_migrations
+            migration_context.migrations_status.collect(&:first).count("down")
```
syncrou: If the `entrypoint` (or `container.spec.args -c ...`) was changed to `bundle exec rake db:migrate && bundle exec rails s`
This `pending_migrations` check would not be needed.

However if the container startup command is not changed then I would like to keep this here as we've already ( in Catalog ) experienced deployments that deploy but are not usable due to pending migrations.

bdunne: I'm not sure I understand this use case.  What handles migrating and seeding?

lindgrenj6: The main use-case here is that we (as in Catalog) have had a few deployments appear to go smoothly, but then a piece of functionality is broken (icons several times in our case) due to a failed migration. 

This would just add a little piece of troubleshooting just in case a migration failed and the application _appears_ to be running normally, but something is degraded somehow.

lindgrenj6:
@bdunne the goal is to pull the fields into kibana or something - that way we can keep track of the state over time etc.

bdunne:
> @bdunne the goal is to pull the fields into kibana or something - that way we can keep track of the state over time etc.

- To get this data into Kibana you would need to log it.
- If there's a metric that you want to keep track over time, it should probably be registered in the Prometheus export server

lib/insights/api/common/status.rb
bdunne: Shouldn't the new api container fail to start if the migration fails?

lindgrenj6: Currently it doesn't fail - we could change that though if we changed the entrypoint for catalog/others to `bundle exec rake db:migrate && bundle exec rails server`, but it just continues currently, as seen here:
https://github.com/RedHatInsights/catalog-api/blob/f1fb5bf0b288c4543029926780f97490bf0219aa/docker-assets/run_rails_server#L27-L28

Is that a change we'd be comfortable with doing? We could remove the migration checking stuff from this if we did it. @syncrou @carbonin @gmcculloug

bdunne: @lindgrenj6 Let's move the migration discussion to https://github.com/RedHatInsights/insights-api-common-rails/issues/146

bdunne:
If we split the migration issues out of this and just focus on the liveness probe, maybe we should just return HEAD 200?  After reading [this blog article](https://blog.openshift.com/liveness-and-readiness-probes/) it sounds like our liveness probe shouldn't even check the connection to the database.  What do you think?

lindgrenj6:
@bdunne  Yes, I'm ok with removing most (if not all) of that logic in the controller as long as we make the pod impossible to start if the database migrations fail.

lindgrenj6:
@bdunne @carbonin I pulled out the db migration stuff - opened a PR on the catalog side to change the behavior of the container to _not_ start up if the database migrations fail. https://github.com/RedHatInsights/catalog-api/pull/563

lib/insights/api/common/status.rb
```diff
@@ -0,0 +1,17 @@
+module Insights
+  module API
+    module Common
+      module Status
+        module Api
+          def health
+            if PG::Connection.ping(ENV['DATABASE_URL']) == PG::Connection::PQPING_OK
+              head :ok
+            else
+              head :internal_server_error
```
bdunne: Won't this cause the app pod to be killed and redeployed if the database is down?

carbonin: Yes, but is that a problem? Also I think that is going to be the least likely failure case.

bdunne: Restarting the pod when the database is down is not going to fix the problem

carbonin: But there could be other potential issues that cause us to be unable to connect to the database which would be solved by a restart. Maybe another thread is in some loop that built up too many connections to the database, maybe the certs changed and we need a reboot to pick up the new ssl config map or even the creds could have changed and the pod isn't restarted when the secret is updated.

These all seem more likely than RDS actually going down. Additionally, if the database is down, then restarting the pod doesn't really seem worse than every request failing and the pod being up. Every request is failing either way.

carbonin: Or maybe we hit some mysterious routing issue in openshift that prevents us from accessing a database which is up. (Not that that has ever happened :trollface:)

lindgrenj6: I'm all for failing _very_ loudly when something else is down, especially since we'd be throwing 500s anyway.

PR #147, opened 2020-01-14
Don't forget to log the error too...

bdunne:
Without this change, errors are not logged and it's very difficult to get the stack trace.


mkanoor:
👍 We need this feature badly in catalog-api and added the logging there and re-raising the error

PR #149, opened 2020-01-16
Allow filtering of tag join records based on tag attributes

bdunne:
- Switch to `arel_attribute(key)` to enable use of `virtual_delegates`
- Allow filtering on extra attributes

@kbrock Please review

### DIFF
--- a/lib/insights/api/common/filter.rb
+++ b/lib/insights/api/common/filter.rb
@@ -5,13 +5,14 @@ module Insights
         INTEGER_COMPARISON_KEYWORDS = ["eq", "gt", "gte", "lt", "lte", "nil", "not_nil"].freeze
         STRING_COMPARISON_KEYWORDS  = ["contains", "contains_i", "eq", "eq_i", "starts_with", "starts_with_i", "ends_with", "ends_with_i", "nil", "not_nil"].freeze
 
-        attr_reader :apply, :arel_table, :api_doc_definition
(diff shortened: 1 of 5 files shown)


lib/insights/api/common/filter.rb
```diff
@@ -5,13 +5,15 @@ class Filter
...
 
-        attr_reader :apply, :arel_table, :api_doc_definition
+        attr_reader :apply, :arel_table, :api_doc_definition, :extra_filterable_attributes, :model
 
-        def initialize(model, raw_filter, api_doc_definition)
-          self.query          = model
-          @arel_table         = model.arel_table
-          @raw_filter         = raw_filter
-          @api_doc_definition = api_doc_definition
+        def initialize(model, raw_filter, api_doc_definition, extra_filterable_attributes = {})
```
kbrock: would you add yard doc to either the `initialize` or the `attr_reader` to define these attributes.

Specifically, what is `extra_filterable_attributes`

In there it can state that the keys are a `String`.

abellotti:
@bdunne looks good, may add the specs for the tag filtering.  Thanks!!

lib/insights/api/common/filter.rb
bdunne: Done

bdunne:
> @bdunne looks good, may add the specs for the tag filtering. Thanks!!

I added tests for the undocumented attributes.  I'll add tests for the `virtual_delegates` in the other PR.

PR #151, opened 2020-01-24
Set the limit to be the max specified in the openapi.json

mkanoor:
https://github.com/RedHatInsights/insights-rbac/blob/e8816dc36041945a62b674203b9062abfcf9ae6d/docs/source/specs/openapi.json#L1491

The default of 10 causes to many back and forth between the RBAC service and the clients.

### DIFF
--- a/lib/insights/api/common/rbac/service.rb
+++ b/lib/insights/api/common/rbac/service.rb
@@ -15,7 +15,7 @@ module Insights
 
           def self.paginate(obj, method, pagination_options, *method_args)
             Enumerator.new do |enum|
-              opts = { :limit => 10, :offset => 0 }.merge(pagination_options)
+              opts = { :limit => 1000, :offset => 0 }.merge(pagination_options)
               count = nil
               fetched = 0
               begin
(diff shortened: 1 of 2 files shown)


mkanoor:
@syncrou Please review

mkanoor:
The generated client gem doesn't have a constant to store this value it is hard coded like here
https://github.com/RedHatInsights/insights-rbac-api-client-ruby/blob/d63a05de04c910ed87292e3ff2d65a5aaa09f0fd/lib/rbac-api-client/api/principal_api.rb#L43
from the openapi.json when the client is generated

lib/insights/api/common/rbac/service.rb
```diff
@@ -15,7 +15,7 @@ def self.call(klass)
 
           def self.paginate(obj, method, pagination_options, *method_args)
             Enumerator.new do |enum|
-              opts = { :limit => 10, :offset => 0 }.merge(pagination_options)
+              opts = { :limit => 1000, :offset => 0 }.merge(pagination_options)
```
gmcculloug: I would agree that 10 is too small but something tells me using this max value is not correct either.  While it will reduce the number of times we have to call the RBAC service it will take time to respond and parse the results.

It would be useful to discuss the bigger picture of what problem we are trying to solve.  How is this call being used and are there other possible solutions.

cc @syncrou

mkanoor:
asking the RBAC team if they can add a feature to filter groups based on one or more uuids. Hence closing

PR #155, opened 2020-02-04
[WIP] Prefer standard `where` over AREL

bdunne:


lib/insights/api/common/filter.rb
```diff
@@ -111,55 +111,55 @@ def integer(k, val)
...
-          self.query = query.where(key => value)
+          self.query = query.where(key => Array(value).collect { |i| ActiveRecord::Base.sanitize_sql(i) })
         end
 
         def comparator_eq_i(key, value)
-          values = Array(value).map { |v| query.sanitize_sql_like(v.downcase) }
+          values = Array(value).collect { |i| ActiveRecord::Base.sanitize_sql(i.downcase) }
 
-          self.query = query.where(arel_table[key].lower.matches_any(values))
+          self.query = query.where("lower(#{key}) IN (?)", values)
```
carbonin: If we're interpolating the key here, should we use something like this to sanitize it? https://www.rubydoc.info/gems/pg/PG/Connection#quote_ident-class_method

(this applies generally to this PR)

lib/insights/api/common/filter.rb
```diff
@@ -111,55 +111,55 @@ def integer(k, val)
...
         end
 
         def comparator_gt(key, value)
-          self.query = query.where(arel_table[key].gt(value))
+          self.query = query.where("#{key} > ?", value)
```
carbonin: Found this which explains an interesting side-effect of this change we might want to consider https://stackoverflow.com/a/35186805

Specifically that we lose the table qualification here so if we're joining across tables that have the same key as a column we will run into errors.

lib/insights/api/common/filter.rb
bdunne: The only things that ever get this far are documented attributes.  Do you think we still need to quote them?

lib/insights/api/common/filter.rb
bdunne: interesting

bdunne:
It doesn't look like this is going to work with the tag filtering I need to do

carbonin:
So just as a recap, did this fix the issue with the acts_as_tenant gem?
Also curious what about the filtering would make a difference between these implementations.

PR #156, opened 2020-02-07
Rename base_path to server_base_path

syncrou:
This method name tripped me up for a bit as
I was thinking it was reading the base path off of disk.
I renamed it to help show that it pulls the value directly from the `servers` key
of the open api doc.


bdunne:
Do you mind updating the open_api/generator for consistency?

syncrou:
@bdunne I modified the `open_api/generator` method name to match as requested.

PR #157, opened 2020-02-13
Change BYPASS_RBAC to only check for the value "true"

gmcculloug:
Change BYPASS_RBAC to only check for the value "true".

Based on changes in https://github.com/RedHatInsights/e2e-deploy/pull/1062 the `BYPASS_RBAC` setting was changed from not being set to defaulting to `false`.

The original logic was setup to bypass RBAC if the environment variable exists, regardless of the value.

### DIFF
--- a/spec/lib/insights/api/common/rbac/access_spec.rb
+++ b/spec/lib/insights/api/common/rbac/access_spec.rb
@@ -33,8 +33,8 @@ describe Insights::API::Common::RBAC::Access do
     expect(described_class.enabled?).to be_truthy
   end
 
-  it "rbac is enabled by default" do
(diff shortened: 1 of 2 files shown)


spec/lib/insights/api/common/rbac/access_spec.rb
```diff
@@ -34,7 +34,7 @@
   end
 
   it "rbac is enabled by default" do
```
bdunne: This test should have a better description like "rbac is disabled with ENV var BYPASS_RBAC=true"

bdunne:
> Limit BYPASS_RBAC values to the environment variable strings "1" and "true".

The description and title should be updated because it's just "true", right? (Which I think is best)

PR #159, opened 2020-02-17
Filter by association attributes

abellotti:
- [x] Depends on https://github.com/RedHatInsights/insights-api-common-rails/pull/149

Adding support for filtering by association.

Enhanced filter usage here: https://github.com/RedHatInsights/insights-api-common-rails/pull/170

### DIFF
--- a/lib/insights/api/common/filter.rb
+++ b/lib/insights/api/common/filter.rb
@@ -4,6 +4,7 @@ module Insights
       class Filter
         INTEGER_COMPARISON_KEYWORDS = ["eq", "gt", "gte", "lt", "lte", "nil", "not_nil"].freeze
         STRING_COMPARISON_KEYWORDS  = ["contains", "contains_i", "eq", "eq_i", "starts_with", "starts_with_i", "ends_with", "ends_with_i", "nil", "not_nil"].freeze
+        ALL_COMPARISON_KEYWORDS     = (INTEGER_COMPARISON_KEYWORDS + STRING_COMPARISON_KEYWORDS).uniq.freeze
 
         attr_reader :apply, :arel_table, :api_doc_definition, :extra_filterable_attributes, :model
 
@@ -23,16 +24,15 @@ module Insights
         # A new Filter object, call #apply to get the filtered set of results.
         #
         def initialize(model, raw_filter, api_doc_definition, extra_filterable_attributes = {})
-          @raw_filter = raw_filter
-          self.query  = if filter_associations.present?
-                          model.left_outer_joins(filter_associations)
-                        else
-                          model
-                        end
+          @raw_filter                  = raw_filter
           @api_doc_definition          = api_doc_definition
           @arel_table                  = model.arel_table
           @extra_filterable_attributes = extra_filterable_attributes
           @model                       = model
+        end
+
+        def query
+          @query ||= filter_associations.present? ? model.left_outer_joins(filter_associations) : model
         end
 
         def apply
@@ -67,12 +67,11 @@ module Insights
           return filter unless filter.kind_of?(Hash) || filter.kind_of?(ActionController::Parameters)
 
           filter = Hash(filter.permit!) if filter.kind_of?(ActionController::Parameters)
-          filter_operators = (INTEGER_COMPARISON_KEYWORDS + STRING_COMPARISON_KEYWORDS).uniq
 
           filter.each do |ak, av|
             if av.kind_of?(Hash)
               av.each do |atk, atv|
-                if !filter_operators.include?(atk)
+                if !ALL_COMPARISON_KEYWORDS.include?(atk)
                   result["#{ak}.#{atk}"] = atv
                 else
                   result[ak] = av
@@ -99,7 +98,6 @@ module Insights
 
         private
 
-        attr_accessor :query
         delegate(:arel_attribute, :to => :model)
 
         class Error < ArgumentError; end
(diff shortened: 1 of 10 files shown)


spec/requests/api/v1.0/filter_spec.rb
```diff
@@ -101,6 +101,28 @@ def expect_success_ordered_objects(collection, query, results)
...
+    let!(:rhev)      { SourceType.create(:name => "rhev_sample",      :product_name => "RedHat Virtualization", :vendor => "redhat") }
+    let!(:openstack) { SourceType.create(:name => "openstack_sample", :product_name => "OpenStack",             :vendor => "redhat") }
+    let!(:openshift) { SourceType.create(:name => "openshift_sample", :product_name => "OpenShift",             :vendor => "redhat") }
+
+    let!(:source_rhev)       { Source.create!(:name => "rhev_source_sample",      :tenant => tenant, :source_type => rhev)      }
+    let!(:source_openstack)  { Source.create!(:name => "openstack_source_sample", :tenant => tenant, :source_type => openstack) }
+    let!(:source_openshift)  { Source.create!(:name => "openshift_source_sample", :tenant => tenant, :source_type => openshift) }
+
+    it("succeeds on a single association attribute value") do
+      expect_success("sources", "filter[source_type.name][eq]=#{rhev.name}", source_rhev)
```
abellotti: @bdunne looks like these are the old style, will need to update for the new signature, filter[source_type][name][eq]... to match the GraphQL.

abellotti:
Ok, took a bit of fixin', but I `think` I got it @bdunne, please check the PR and updates of today https://github.com/RedHatInsights/insights-api-common-rails/pull/159/commits/ed911b418e258b71dfdaeb94694a16d0f77b0245 to take the new signature for filtering with association attributes. Thanks!

/cc @gtanzillo

lib/insights/api/common/filter.rb
```diff
@@ -14,21 +14,25 @@ class Filter
         #   An AR model that acts as the base collection to be filtered
         # raw_filter::
         #   The filter from the request query string
-        # api_doc_definition::
-        #   The documented object definition from the OpenAPI doc
+        # api_doc_definitions::
+        #   The documented object definitions from the OpenAPI doc
```
bdunne: How man definitions are there?  Isn't this the definition of a single object (e.g. Cluster)?

lib/insights/api/common/filter.rb
```diff
@@ -14,21 +14,25 @@ class Filter
...
         #
         # == Returns:
         # A new Filter object, call #apply to get the filtered set of results.
         #
-        def initialize(model, raw_filter, api_doc_definition, extra_filterable_attributes = {})
-          self.query                   = model
-          @api_doc_definition          = api_doc_definition
+        def initialize(model, raw_filter, api_doc_definitions, extra_filterable_attributes = {})
+          @raw_filter = raw_filter
+          self.query = if filter_associations.present?
```
bdunne: Having a method call inside `initialize` is not great.  What do you think of splitting it into a method and removing the attr_reader?

```ruby
def query
  @query ||= filter_associations.present? ? model.left_outer_joins(filter_associations) : model
end
```

lib/insights/api/common/filter.rb
bdunne: Also, this breaks API compatability, so it requires a major version.

lib/insights/api/common/filter.rb
```diff
@@ -48,16 +52,72 @@ def apply
...
+        def self.condense(filter)
+          return filter if filter.blank? || !filter.kind_of?(Hash)
+
+          primary_key   = filter.keys.first
+          primary_value = filter[primary_key]
+
+          if filter.keys.size == 1 && primary_value.kind_of?(Hash)
+            secondary_key   = primary_value.keys.first
+            secondary_value = primary_value[secondary_key]
+            filter_operators = (INTEGER_COMPARISON_KEYWORDS + STRING_COMPARISON_KEYWORDS).uniq
```
bdunne: Should this be a constant?

lib/insights/api/common/filter.rb
abellotti: sorry, I had that earlier, but is no longer the case, I'll update the comment.

spec/requests/api/v1.0/filter_spec.rb
bdunne: huh?

lib/insights/api/common/filter.rb
```diff
@@ -14,22 +14,22 @@ class Filter
         #   An AR model that acts as the base collection to be filtered
         # raw_filter::
         #   The filter from the request query string
-        # api_doc_definitions::
-        #   The documented object definitions from the OpenAPI doc
+        # api_doc_definition::
```
bdunne: Oh, this is back to singular now.  Ignore the previous comment

spec/lib/insights/api/common/filter_spec.rb
```diff
@@ -0,0 +1,74 @@
+describe Insights::API::Common::Filter do
+  let(:this) { described_class }
+
+  context "compact_filter method" do
+    it "compacts association filter specifications of string" do
+      parameters  = {"association" => {"attribute" => "value"}}
+      expectation = {"association.attribute" => "value"}
+
+      expect(this.public_send(:compact_filter, parameters)).to(eq(expectation))
```
bdunne: Is there an advantage to doing `this.public_send(:compact_filter, parameters)` rather than `described_class.compact_filter(parameters)`?

spec/requests/api/v1.0/filter_spec.rb
abellotti: confusing sorry, this was from the original prototype, needed to implement the filter/sort_by normalized usage instead.

spec/lib/insights/api/common/filter_spec.rb
abellotti: Nope, you're right. Updating to just described_class.compact_filter(...)

lib/insights/api/common/filter.rb
```diff
@@ -214,63 +212,63 @@ def arel_lower(key)
         def comparator_contains(key, value)
           return value.each { |v| comparator_contains(key, v) } if value.kind_of?(Array)
 
-          self.query = query.where(model_arel_attribute(key).matches("%#{query.sanitize_sql_like(value)}%", nil, true))
+          @query = query.where(model_arel_attribute(key).matches("%#{query.sanitize_sql_like(value)}%", nil, true))
```
bdunne: Should we also define a setter method so that these don't need to change?
```ruby
def query=(new)
  @query = new
end
```

abellotti: 👍 with one minor update, Rubocop preferred having the attr_writer instead, works the same and we don't need to change the self.query = ...

PR #160, opened 2020-02-20
[WIP] Moving the sort_by parameter specification as an OpenAPI deep object.

abellotti:
Moving the sort_by parameter specification as an OpenAPI deep object.

- This is to get around an issue with the OpenAPI Client Generator not properly
  handling the OneOf spec directive.

- Needed to move the validation of the sort_by parameter into the PaginatedResponse
  class where it is used.

Solves: TPINVTRY-864

Related PRs:
- https://github.com/RedHatInsights/topological_inventory-api/pull/282
- https://github.com/RedHatInsights/sources-api/pull/188


mkanoor:
👍 Looks good
went from
```
opts = {
  limit: 100, # Integer | The numbers of items to return per page.
  offset: 0, # Integer | The number of items to skip before starting to collect the result set.
  filter: nil, # Object | Filter for querying collections.
  sort_by: TopologicalInventoryApiClient::OneOfstringarray.new # OneOfstringarray | The list of attribute and order to sort the result set by.
}
``` 
to
```
opts = {
  limit: 100, # Integer | The numbers of items to return per page.
  offset: 0, # Integer | The number of items to skip before starting to collect the result set.
  filter: nil, # Object | Filter for querying collections.
  sort_by: nil # Object | The list of attribute and order to sort the result set by.
}
```

lib/insights/api/common/paginated_response.rb
```diff
@@ -102,6 +103,14 @@ def sort_by_options(model)
             end
           end
         end
+
+        def validate_sort_by
+          Array(sort_by).each { |key| validate_sort_by_directive(key) }
```
bdunne: Should this be `sort_by.each_key`?

abellotti: we have an array, either from the single sort_by, the the multiple sort_by[] query parameters, I thought each_key is for hashes, so we really can't use that here.

bdunne: That explains why this confused me :smile: 
`type: object` in the OpenAPI doc implies that a hash is coming in, right?  But I see that the specs are still testing with a string or an array as the input.

abellotti: correct, didn't want to break current interface. so yes, deep object, but handing string, arrays, and shortly hash also (once we bring in the association sort_by).

bdunne: If we're changing the type to "object" then we're breaking the interface and it doesn't make sense to allow strings and arrays.

bdunne: Also, we have to figure out how to implement this in a way that we don't break older versions of the API running on the same rails server.

abellotti: I think if we want pure objects for sort_by, then we'd need to bump up the version of topo/sources/etc to 3.0 per semver because it is a breaking change.  The old code would stick around in 2.0 classes and we'll do the object way in the 3.0 source.

bdunne: Yes, this change will require a major version bump for consumers.  But we still need a way to support the old API versions, so maybe we need to move this change into a `paginated_response_v2`?

abellotti:
Closing in preference for #165

PR #161, opened 2020-02-25
According to the IPP auth_type is in identity

agrare:
`IPP-14 Red Hat Insights User Identity` specifies that the auth_type is under the identity section not the internal section.


agrare:
@abellotti can you take a look?  Is this correct or is the IPP wrong?

bdunne:
Unfortunately the platform docs don't go into enough detail here.

PR #162, opened 2020-02-25
Allow machine credentials access to the API

agrare:
When a machine credential connects to 3scale there is no user key in the headers but rather a system key.

Depends on: https://github.com/RedHatInsights/insights-api-common-rails/pull/162
Relates to JIRA: https://projects.engineering.redhat.com/browse/TPINVTRY-870

### DIFF
--- a/spec/lib/insights/api/common/request_spec.rb
+++ b/spec/lib/insights/api/common/request_spec.rb
@@ -51,6 +51,18 @@ describe Insights::API::Common::Request do
       expect(@instance.user).to be_a(Insights::API::Common::User)
     end
 
+    it "#tenant" do
+      expect(@instance.tenant).to eq(default_account_number)
+    end
+
+    it "#system" do
+      expect(@instance.system).to be_nil
+    end
+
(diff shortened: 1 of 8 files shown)


spec/lib/insights/api/common/request_spec.rb
```diff
@@ -93,6 +105,68 @@
...
+
+    it "#system" do
+      expect(@instance.system).to be_a(Insights::API::Common::System)
+    end
+
+    it "#auth_type" do
+      expect(@instance.auth_type).to eq("cert-auth")
+    end
+
+    it "#user" do
```
bdunne: Duplicate of https://github.com/RedHatInsights/insights-api-common-rails/pull/162/files#diff-b602271974e4016385b4ea6a117dd8c4R141

agrare: :+1: deleted

bdunne:
I released [v3.5.0](https://rubygems.org/gems/insights-api-common/versions/3.5.0) with this change

PR #164, opened 2020-02-27
Move group validation to separate class and add specs

eclarizio:
As discussed in Catalog PR https://github.com/RedHatInsights/catalog-api/pull/596, this change should probably live here since it will be used by both Catalog and Approval.

@mkanoor @syncrou @gmcculloug Please Review.

### DIFF
--- a/lib/insights/api/common/rbac/validate_groups.rb
+++ b/lib/insights/api/common/rbac/validate_groups.rb
@@ -1,0 +1,24 @@
+module Insights
+  module API
+    module Common
+      module RBAC
+        class ValidateGroups
+          def initialize(group_uuids)
+            @group_uuids = group_uuids
+          end
+
+          def process
+            return unless Insights::API::Common::RBAC::Access.enabled?
+
+            Service.call(RBACApiClient::GroupApi) do |api|
+              uuids = SortedSet.new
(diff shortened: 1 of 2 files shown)


lib/insights/api/common/rbac/utilities.rb
```diff
@@ -3,15 +3,6 @@ module API
     module Common
       module RBAC
         module Utilities
-          def validate_groups
```
gmcculloug: @mkanoor You mentioned the other methods in the module are no longer used, correct?  Do we remove this entire file?

eclarizio: I think even if we do, it probably doesn't belong in the scope of this PR (as small of a change as that is)

mkanoor: @gmcculloug We can do that in a separate PR, we also need @hsong-rh to update the Approval code to use this new class instead of the old mixin and we can then have a PR to delete it for good.

mkanoor: @eclarizio Can we not touch this file for now and remove it in a different PR, if we remove it now it will break Approval. Once this is merged we would have to wait for @hsong-rh and then get rid of this file entirely.

eclarizio: Sure, that's fine.

PR #167, opened 2020-03-04
Sort by association attribute and count

abellotti:
- Adding support for sorting by association attribute and association count using the object based specification for the sort_by parameter.

_Note_: with sort_by coming in as hash parameters, GraphQL does not like periods or @ in parameter names, periods are gone, but the **@ count** needed to be changed, went with **__count**.

Fix for: TPINVTRY-770
Depends on: https://github.com/RedHatInsights/insights-api-common-rails/pull/165

### DIFF
--- a/lib/insights/api/common/paginated_response_v2.rb
+++ b/lib/insights/api/common/paginated_response_v2.rb
@@ -2,7 +2,6 @@ module Insights
   module API
     module Common
       class PaginatedResponseV2 < PaginatedResponse
-
         # GraphQL name regex: /[_A-Za-z][_0-9A-Za-z]*/
         ASSOCIATION_COUNT_ATTR = "__count".freeze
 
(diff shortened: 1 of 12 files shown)


lib/insights/api/common/paginated_response_v2.rb
```diff
@@ -2,35 +2,122 @@ module Insights
...
+        end
+
+        # Condenses parameter values for handling multi-level associations
+        # and returns an array of key, value pairs.
+        #
+        # Input:  { "association" => { "attribute" => "value" }, "direct_attribute" = "value2" }
+        # Output: { "association.attribute" => "value", "direct_attribute" => "value2" }
+        #
+        # Input:  { "association" => { "attribute" => "value" }, "association2" => "attribute2" = "value2" }
+        # Output: { "association.attribute" => "value", "association2.attribute2" => "value2" }
```
bdunne: From the looks of it, this will return an Array of Arrays, not a hash.  Can you turn these comments into unit tests for this method?

bdunne: Also, it doesn't look like it would cover `{"association" => {"attribute" => "value", "attribute2" = "value2"}`

abellotti: thanks for finding this.  fixing this exposed an issue when referencing same named attribute from class and association class.   will be pushing an update shortly. 👍

spec/lib/insights/api/common/paginated_response_v2_spec.rb
```diff
@@ -0,0 +1,43 @@
+describe Insights::API::Common::PaginatedResponseV2 do
+  context "compact_parameter method" do
```
bdunne: :+1: Nice!

PR #170, opened 2020-03-13
Updated README.md for object syntax of sort_by and filtering and sorting by association

abellotti:
Updated the README.md to mention the new object type syntax for sort_by parameter as well as the filtering and sorting by association attribute as well as sorting by association count.

Dependencies:

- [x] https://github.com/RedHatInsights/insights-api-common-rails/pull/167
- [x] https://github.com/RedHatInsights/insights-api-common-rails/pull/159

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,23 +69,56 @@ After implementing filtering in your application, this is the way to filter via 
 
 Sorting query results is controlled via the _sort_by_ query parameter. The _sort_by_ parameter is available for both REST API and GraphQL requests.
 
-The syntax for the _sort_by_ parameter supports:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -69,23 +69,56 @@ After implementing filtering in your application, this is the way to filter via
...
+
+#### Filtering and Sorting by Association attribute
+
+Requests can also be filtered by assocation attribute and sorted by association attribute and count in addition to the direct attribute specified as in the above examples.
+
+Single level association can be specified as follows:
+
+##### Filter by association attribute:
+
+| Query Parameter | Ruby Client Parameter <br> **GraphQL filter: Parameter** |
```
bdunne, line 101: Should these tables be 3 columns like the one above?

abellotti: I think that will look better, lemme give that a try 👍

abellotti: ok, pushed an update. Decided to normalize all tables with the 3 columns, much nicer to see. Thanks!!

PR #171, opened 2020-03-19
RBAC - Added support for scopes in permissions

mkanoor:
In the RBAC resource definition we are adding scope to
differentiate between admin, groups and users

Sample Usage by Approval

```
x = Insights::API::Common::RBAC::Access.new
my_scopes = x.scopes('workflows','read')
if my_scopes.include?('admin')
    # handle admin scope 
    # Worfklows.all
elsif my_scopes.includes?('group')
   # Handle Group Scope
elsif my_scopes.includes?('user')
   # Handle User Scope
else
   # Nothing Accessible
end
```
Sample Usage by Catalog API
```
x = Insights::API::Common::RBAC::Access.new('catalog, approval')
my_scopes = x.scopes('portfolios','read')
if my_scopes.include?('admin')
    # handle admin scope 
    Portfolios.all
elsif my_scopes.includes?('group')
   # Handle Group Scope
elsif my_scopes.includes?('user')
   # Handle User Scope
else
   # Nothing Accessible
end

# Somewhere else when doing Set Approval
my_scopes = x.scopes('workflows', 'read',  'approval')
my_tags_scopes = x.scopes('taglinks', 'write',  'approval')
if my_scopes.count > 0 && my_tags_scopes.count > 0
    # User can set approval on Portfolio
end

```


lib/insights/api/common/rbac/access.rb
```diff
@@ -25,6 +25,23 @@ def accessible?
             @acl.any?
           end
 
+          def is_accessible?(app_name, resource, verb)
```
hsong-rh: For following methods, may we set default value `ENV["APP_NAME"]` for `app_name`?  They might be called multiple times in each app.  Add a check on `ENV["APP_NAME"].nil?` ?

mkanoor: @hsong-rh I thought about that but it runs into problems about where the argument is

is_accessible?(resource, verb, app_name = ENV['APP_NAME'])

but most of the permissions look like

catalog:portfolios:read

Where app_name comes first.

In the case of catalog we have to check the permission for catalog and approval so it has to be passed in all the time.

lib/insights/api/common/rbac/access.rb
```diff
@@ -25,6 +25,23 @@ def accessible?
...
+          def admin_scope?(app_name, resource, verb)
+            scope?(app_name, resource, verb, "admin")
+          end
+
+          def group_scope?(app_name, resource, verb)
+            scope?(app_name, resource, verb, "group")
+          end
+
+          def user_scope?(app_name, resource, verb)
+            scope?(app_name, resource, verb, "user")
```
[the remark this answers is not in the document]
mkanoor: @bzwei Why not be explicit about it, there is no harm in saying that certain resources are user specific.

lib/insights/api/common/rbac/access.rb
```diff
@@ -25,6 +25,23 @@ def accessible?
             @acl.any?
           end
 
+          def is_accessible?(app_name, resource, verb)
+            regexp = create_regexp(app_name, resource, verb)
+            @acl.any? { |item| regexp.match?(item.permission) }
+          end
+
+          def admin_scope?(app_name, resource, verb)
+            scope?(app_name, resource, verb, "admin")
```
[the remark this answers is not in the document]
mkanoor: @bzwei yes, we have not been using the id == '*' approach for a while now. We had that in our first implementation. We will remove those functions once we have everyone at the latest version.

lib/insights/api/common/rbac/access.rb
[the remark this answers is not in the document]
mkanoor: @bzwei The is_accessible is the new method which will replace the accessible? method. The current accessible? method doesn't take any arguments. The is_accessible? takes the app name, resource, verb.

lib/insights/api/common/rbac/access.rb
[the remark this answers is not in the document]
bzwei: Yes we can set explicitly for user scope. What I suggest is the default should be user scope if there is no resourceDefinition for it. Therefore we can have a much clean permission definition for regular users.

lib/insights/api/common/rbac/access.rb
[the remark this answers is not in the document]
bzwei: Since the gem is version controlled, we can safely replace the existing `accessible?` method with the new implementation. Any app that uses this gem will update the use of `accessible?` and bump up the gem version number. We don't need to have another round to rename `is_accessible?` to `accessible?`.

lib/insights/api/common/rbac/access.rb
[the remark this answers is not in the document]
bzwei: Again, the gem is version controlled, we can safely remove it now without worrying someone is not using the latest version.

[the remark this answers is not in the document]
bzwei: Sorry, depend on how the gem version is specified in Gemfile, we do need to maintain backward compatibility. But if we are sure nobody is using it we can remove it now.

lib/insights/api/common/rbac/access.rb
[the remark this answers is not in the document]
bzwei: Sorry, depend on how the gem version is specified in Gemfile, we do need to maintain backward compatibility.
You may want to consider to have `def accessible?(app_name = nil, resource = nil, verb = nil)` for a while for backward compatibility purpose. And in the future remove the default values for arguments. This way at least both catalog-api and approval-api does not need to switch from `is_accessible?` to `accessible?`.

mkanoor:
@bzwei @hsong-rh Please review removed old methods and cleaned up specs.

lib/insights/api/common/rbac/access.rb
```diff
@@ -5,32 +5,32 @@ module RBAC
...
-                @regexp.match?(item.permission)
-              end
+              @acl ||= Service.paginate(api, :get_principal_access, {:limit => DEFAULT_LIMIT}, @app_name).to_a
             end
             self
           end
 
-          def accessible?
-            @acl.any?
+          def accessible?(resource, verb, app_name = ENV['APP_NAME'])
```
bzwei: `app_name` is already given by `initialize`

gmcculloug: Right, maybe these should be `def accessible?(resource, verb, app_name = @app_name)`

mkanoor: @gmcculloug When we fetch permissions for catalog and approval in a single RBAC call we might have to set @app_name = '*'  or @app_name = '\*' or @app_name = ""

But when you call this function you are trying to fetch it for a specific  app, so the @app_name is not a good default in this case. So the app_name = ENV['app_name'] is a better default and it sets the default to the current app that is using it.

mkanoor: To clear the confusion I have changed the app_name in the constructor to app_name_filter since it can be '*'

bzwei: @mkanoor I am confused. When `@acl` is created, they are already limited by `@app_name_filter`. What do you get if here you `app_name = '*'`?

Or is it the opposite that you want to set `@app_name_filter = '*'` first and narrow it down to `app_name = 'catalog'` here? By using '*', do you really want to fetch permissions from ALL apps? The list will be very long. It might be better to call the method twice for two selected apps.

mkanoor: @bzwei 
The RBAC service stores all the permissions for a user across applications. There are many applications that use RBAC and a single user can have permissions from many applications.

The get_principal_access call to the RBAC service gives you back these access/permissions. This API takes a filter called application_name listed here


https://github.com/RedHatInsights/insights-rbac-api-client-ruby/blob/master/docs/AccessApi.md#get_principal_access

The first argument to get principal access call is the name of the application to filter the access by so for a single application you can fetch your access by using the ENV['APP_NAME'] which is a pretty good default for services that only need to fetch access list for a single app e.g approval.

For apps that need access for 2 apps like catalog, we need access for approval and catalog we would have to make either 2 calls or make a single call with '*' or '' and then locally filter out the permissions we need.

Approach 1:

  x = get_principal_access('catalog',..)
  y = get_principal_access('approval',..)

Approach 2:
z = get_principal_access('*',...)

Or
 Approach 3:
a = get_principal_access('catalog, approval',...)


the get_principal_access call is expensive so its beneficial to make a single call.

Once you have the access available locally we cache that for the single API call since we might have to traverse the result multiple times.

So the admin_scope?, group_scope?, user_scope? calls are dealing with the locally available data so we don't have to go back to RBAC on a per call basis.

So when we initialize a instance of the Access class we are asking for the filter which can be

'approval'
'approval, catalog',
''
'*'

This is not the value that you would use when you call the admin_scope?, group_scope?, user_scope?, that has to be specific for the app that you are interested in.

As far as apps that deal with a single service the app_filter is the same as the ENV['APP_NAME'] and you don't have to add an extra argument

But for apps that have to deal with permissions for 2 different apps the app_name argument becomes important and it cannot default to the app_name_filter.

bzwei: My concern is to get all back through '*'. Since we can select multiple apps `a = get_principal_access('catalog, approval',...)`, then yes, this is good. Thanks for the clarification.

mkanoor:
@bzwei 
We have a ticket opened with the RBAC team https://projects.engineering.redhat.com/browse/RHCLOUD-5269
Which might give us more clarity on what the app_name filter would look like

'catalog,approval'
'*'
''

or something else

lib/insights/api/common/rbac/access.rb
gmcculloug: My concern is if someone initializes the object with `app_name_filter` set to a single app_name that is not the same as `ENV['APP_NAME']` then they are required to pass `app_name` to every method call, which seems odd.

Does that seem like a reasonable use case to handle?

I think we could support this in the initialize method by allowing a user to pass the `app_name` as a second optional parameter, and if that is nil (the default) we set it to `app_name_filter` if it does not include a separator or wildcard character.  Otherwise, it defaults to `ENV['APP_NAME']` .

Something like this:
```ruby
def initialize(app_name_filter = ENV["APP_NAME"], app_name = nil)
  @app_name_filter = app_name_filter

  @app_name = if app_name.present?
                app_name
              elseif !app_name_filter.include?('*') and !app_name_filter.include?(',')
                app_name_filter
              else
                ENV["APP_NAME"]
              end
end
```

Probably a better way to do this, but trying to get the idea across.

mkanoor:
@gmcculloug I am not sure adding an extra parameter helps anything here for a normal use case say for approval which only ever works with only its objects the code would look like

```
x = Insights::API::Common::RBAC::Access.new
x.accessible?('workflows','read')
x.admin_scope?('workflows','read')
```

For a case like catalog which needs to get access from catalog and approval and maybe topology down the road

```
x = Insights::API::Common::RBAC::Access.new('catalog, approval')
x.accessible?('portfolio','read')
x.admin_scope?('portfolio','read')

x.accessible?('workflow','read','approval')
```

You need to pass in the extra argument only when you are trying to get permission check for objects **which are not** in your app, When you need to check for permissions **in your app** you dont have to pass in the extra argument. It couldn't be much simpler than that. Adding extra argument in the initializer has no value since the app name is needed when you make the check. Only when you need to check for permissions for objects not in your own app.

mkanoor:
I changed this to WIP temporarily because the caller might see a big performance improvement if it collected all the scopes in a single call and then did some logic locally.

e.g
```
x = Insights::API::Common::RBAC::Access.new
if x.admin_scope?('workflows', 'update')
   # Do Admin Stuff
else if x.group_scope?('workflows','update')
  # Do Group Stuff
else if x.user_scope?('workflows,'update')
  # Do User Stuff
else
     puts "Not Accessible"
end
```

versus Collecting all scopes in a single call so you dont have to traverse the array multiple times.
```
x = Insights::API::Common::RBAC::Access.new
current_scopes = x.scopes('workflows', 'update')
if current_scope.includes?('admin')
   # Do Admin Stuff
else if current_scope.includes?('group')
   # Do Group Stuff
else if current_scope.includes?('user')
   # Do User Stuff
else
   puts "Not Accessible"
end
```

mkanoor:
@hsong-rh @bzwei 
Please review the example usages. I have added a new method called scopes since it will make it easier to make a single pass and collect all the scopes for an object.

The use case is

A user can be part of all these groups at the same time
   - Approval Administrator
   - Approval Approver
   - Approval User

Or the user could be part of 2 groups
   - Approval Administrator
   - Approval Approver

So the scopes method allows you to collect all the scopes with a single call and then have the caller decide the order of precedence. Admin will be checked first then group and then user.

gmcculloug:
@mkanoor I like the new approach as it does not require this code to know the scope values (admin, group, user) and the app has more control over how it accesses the data.  👍

lib/insights/api/common/rbac/access.rb
gmcculloug: I don't have a real use-case for this at the moment, so willing to ignore this for now and address later if we find a need, at which time we'll have more information to drive the changes.

In most cases the app is going to be just checking it's own permissions so this makes sense as is.

lib/insights/api/common/rbac/access.rb
```diff
@@ -39,26 +52,33 @@ def self.enabled?
...
             end
           end
 
-          def owner_scope_filter?
-            @acl.any? do |item|
-              item.resource_definitions.any? do |rd|
-                rd.attribute_filter.key == 'owner' &&
-                  rd.attribute_filter.operation == 'equal' &&
-                  rd.attribute_filter.value == '{{username}}'
+          def all_scopes(item)
```
bzwei: since each scope can be one of three possible values, and admin > group > user, can a permission in reality have more than one scope defined in the definition?

mkanoor: @bzwei Yes, the attribute filter is an array. Plus the permission is coming from different roles.

Approval Administrator
                   -> approval.workflows.read
                          sope = admin
Approval Approver
                  -> approval:workflows:read
                         scope = group

Once you make a get_principal_access call it dissolves the roles and you are left with
-> approval:workflows:read
                          sope = admin
-> approval:workflows:read
                         scope = group

When the API calls in the above scenario with
scopes('workflows',read') the result will be ['admin','group']

lib/insights/api/common/rbac/access.rb
```diff
@@ -5,32 +5,45 @@ module RBAC
...
 
-          def accessible?
-            @acl.any?
+          def scopes(resource, verb, app_name = ENV['APP_NAME'])
+            regexp = create_regexp(app_name, resource, verb)
+            @acls.each_with_object([]) do |item, memo|
+              if regexp.match?(item.permission)
+                memo << all_scopes(item)
+              end
+            end.flatten
```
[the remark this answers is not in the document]
mkanoor: @bzwei I will remove the duplicates

mkanoor:
@bzwei Please review

lib/insights/api/common/rbac/access.rb
bzwei: '*' is logically a superset of app_name.
```
x = Insights::API::Common::RBAC::Access.new('catalog')
x.accessible?('workflow','read','approval')
```
The above example will always return `false` even if the login user has true read permission to workflow. It is a coding error, but the bug cannot be easily revealed.

PR #172, opened 2020-03-25
Implement group_uuid filtering

mkanoor:
This was added by the RBAC team in
https://github.com/RedHatInsights/insights-rbac/pull/243

### DIFF
--- a/lib/insights/api/common/rbac/validate_groups.rb
+++ b/lib/insights/api/common/rbac/validate_groups.rb
@@ -12,7 +12,7 @@ module Insights
 
             Service.call(RBACApiClient::GroupApi) do |api|
               uuids = SortedSet.new
-              Service.paginate(api, :list_groups, {}).each { |group| uuids << group.uuid }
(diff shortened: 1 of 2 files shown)


lib/insights/api/common/rbac/validate_groups.rb
```diff
@@ -12,7 +12,7 @@ def process
 
             Service.call(RBACApiClient::GroupApi) do |api|
               uuids = SortedSet.new
-              Service.paginate(api, :list_groups, {}).each { |group| uuids << group.uuid }
+              Service.paginate(api, :list_groups, {:uuid => @group_uuids.to_a}).each { |group| uuids << group.uuid }
```
gmcculloug: @mkanoor what is `@group_uuids` at this point?  Does this need `to_a` or `Array(@group_uuids)`

mkanoor: @gmcculloug Its an SortedSet 
https://github.com/RedHatInsights/catalog-api/blob/1b46ac5cd30dbb92ae53376f1bcffaa367f7787b/app/services/catalog/share_resource.rb#L4

Issue #66, opened 2019-05-21
major version redirect should internally redirect and not respond with 302


Fryguy:
If we expect most users to use the major version, then the 302 introduces a lot of chatter overhead, which is unnecessary.

@carbonin

carbonin:
I'm not sure how to accomplish this with what we have currently. We may need to move away from using `redirect` ... @bdunne maybe you'll have better luck with this one.

bdunne:
I expect most users to use a generated client which will point at a minor version of the API (e.g. `/api/vX.Y/`).  The major version redirect is there for convenience and discovery.

Also, I think the major version has to respond with a 302 since we can't publish a static OpenAPI document for the major version because the response payloads will change every time we bump the minor version.
cc @abellotti

PR #173, opened 2020-03-26
Remove POST PATCH as available redirect options

syncrou:
These changes remove the ability to do major version redirects for POST, PATCH REST operations.

POST, PATCH redirection is not recommended and currently is stripping out the data payload when the redirect route is called.

As discussed here: https://softwareengineering.stackexchange.com/questions/99894/why-doesnt-http-have-post-redirect

JIRA: https://issues.redhat.com/browse/RHCLOUD-9248


bdunne:
Wow, the more I read, the worse it gets.  It seems like some clients still don't even honor 307, so I don't think that would help us.
Should we drop all redirects other than GET?
Maybe we should catch the other methods and return an error response telling the user to use the latest minor version for that major version. (and include that version in the response)
cc @abellotti @agrare @Fryguy Thoughts?

Fryguy:
Still think we should just implement #66 and get rid of the redirect nonsense entirely 😉

abellotti:
ya, as per https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html  some agents erroneously change the redirect into a GET, which will surely mess things up, and this is true for 301/302 as well as the 307.

bdunne:
I'm leaning toward leaving this as is since there really shouldn't be any consumers of the major version API.  All consumers should be using a client which would point to a `x.y` version of the API, right?

bdunne:
I'd rather be all or nothing on the major version.  Anyone have better ideas on how to present the major version API without duplicating a ton of routes and controllers?

lindgrenj6:
@syncrou I think we're ready to make this go forward for the UI. 

Can we get a rebase + minor gem release commit on here? Then we should be good to go.

syncrou:
@lindgrenj6 This should be set now. Please take another look when you can.

roliveri (approved):

PR #174, opened 2020-04-01
Use new openapi-parser

mkanoor:
Use the latest openapi-parser it has additional uuid format check.


bdunne:
If the tests pass it should be fine.

PR #176, opened 2020-04-03
Updated the support of association filter to expect operators

abellotti:
While this worked earlier, it was an ambiguous specification when no operators are present, i.e. filter[attr]=value case where eq is not specified.

- Updated the support of association filter so that specification of associations in filter requires the operator to be specified.
- Implied "eq" for values is no longer supported for associations, the "eq" operator should be specified.
- Updated README.md for additional examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ The syntax for the _sort_by_ parameter is as follows:
 
 #### Filtering and Sorting by Association attribute
 
-Requests can also be filtered by assocation attribute and sorted by association attribute and count in addition to the direct attribute specified as in the above examples.
+Requests can also be filtered by assocation attribute and sorted by association attribute and count in addition to the direct attribute specified as in the above examples. When specifying attribute associations, the filter operators must be specified.
 
 Single level association can be specified as follows:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -92,16 +92,16 @@ The syntax for the _sort_by_ parameter is as follows:
...
 
 Single level association can be specified as follows:
 
 ##### Filter by association attribute:
 
 | Query Parameter | Ruby Client Parameter | GraphQL Parameter |
 | --------------- | --------------------- | ----------------- |
-| "?filter[association][name]=reviews" | { :filter => { :association => { :name => "reviews" } } } | filter: { association: { name: "reviews" } } |
 | "?filter[association][name][eq]=reviews" | { :filter => { :association => { :name => { :eq => "reviews" } } } } | filter: { association: { name: { eq: "reviews" } } } |
+| "?filter[source_type][name][eq][]=OpenStack&filter[source_type][name][eq][]=OpenShift" | { :filter => { :source_type => { :name => { :eq => [ "OpenStack", "OpenShift" ] } } } } | filter: { source_type: { name: { eq: [ "OpenStack", "OpenShift" ] } } } |
```
lindgrenj6, line 104: So I get how this is working - why is there an extra `[]` after `[eq]`? I thought that eq was the operator specifier and that would be enough.

bdunne: The `[]`is how you specify an array.  In this case "OpenStack" and "OpenShift".

lindgrenj6: Ahhhhh ok, that makes more sense. that way it ends up with this `filter: { source_type: { name: { eq: [ "OpenStack", "OpenShift" ] } }`

Thanks for the clarification!

bdunne:
I think my only concern is that we're not allowing the implied `eq` comparator if no comparator is given and that is inconsistent between direct attributes and association attributes

abellotti:
If we want to allow the implied eq with associations  (which is what we have today), then the change instead would be to the specs in topo land for [attr][bad_op]=value would be seen as attr.bad_op = value  instead of attr bad_op value, so changing the error message returned. (i.e. bad association attr instead of bad operator bad_op). Let me know @bdunne.

abellotti:
@lindgrenj6 @syncrou any preference here ?

syncrou:
@abellotti so are you saying the implied `eq` for associations is only a spec change in topo?

abellotti:
correct, if we want implied eq, it's already here in common, so I'd close this PR and update specs in topo instead.

abellotti:
any update on this ?  should we stick with the optional eq for associations ?

lindgrenj6:
@abellotti I'm of the opinion maybe we should still allow implied eq, but that's just me. Thoughts @bdunne @syncrou?

abellotti:
Closing this PR in preference for supporting the implied eq on association filters as Sources is already GA'd with this capability.

PR #179, opened 2020-04-13
Use rack to get proper status code

### LINKED ISSUE #178: HTTP errors are not properly translated to status in error doc
Below is an example of an error document where the status is not converted to integer and comes out as the string value of the RAILS http symbol.

```
{
    "errors": [
        {
            "status": "bad_request",
            "detail": "Exceptions::UserError: Group ba403c79-14bb-4a1e-a65c-31ff711d9cfd does not exist"
        },
        {
            "status": 404,
            "detail": "RBACApiClient::ApiError: Error message: the server returns an error\nHTTP status code: 404\nResponse headers: {\"x-powered-by\"=>\"Express\", \"x-fakamai-env\"=>\"ci\", \"x-fakamai-app-hint\"=>\"API\", \"x-fakamai-downstream-host\"=>\"https://api-gateway-ci.5a9f.insights-dev.openshiftapps.com\", \"server\"=>\"openresty/1.13.6.1\", \"date\"=>\"Mon, 13 Apr 2020 17:51:06 GMT\", \"content-type\"=>\"application/json\", \"x-rh-insights-request-id\"=>\"fd74f806f64c4e60a89d00b3d5152750\", \"vary\"=>\"Accept, Origin\", \"allow\"=>\"GET, PUT, PATCH, DELETE, HEAD, OPTIONS\", \"x-frame-options\"=>\"SAMEORIGIN\", \"x-content-type-options\"=>\"nosniff\", \"set-cookie\"=>\"bd2e829019cc5851647126f4f0834941=5040a645e9d4912b5513f4d2f9664fd7; path=/; HttpOnly; Secure\", \"cache-control\"=>\"private\"}\nResponse body: {\"errors\":[{\"detail\":\"Not found.\",\"source\":\"detail\",\"status\":404}]}"
        }
    ]
}
```

mkanoor:
Fixes #178
Uses Rack::Utils.status_code to convert the Rails HTTP error symbols to
numeric values. The status_code can handle integers/strings/symbols and
returns the appropriate numeric value
e.g.

Rack::Utils.status_code(400) => 400
Rack::Utils.status_code("400") => 400
Rack::Utils.status_code(:forbidden) => 403

The JSON API Error Object Sample
https://jsonapi.org/examples/#error-objects-basics

The status value should be
status: the HTTP status code applicable to this problem, expressed as a string value.

### DIFF
--- a/lib/insights/api/common/application_controller_mixins/exception_handling.rb
+++ b/lib/insights/api/common/application_controller_mixins/exception_handling.rb
@@ -11,7 +11,7 @@ module Insights
               errors = Insights::API::Common::ErrorDocument.new.tap do |error_document|
                 exception_list_from(exception).each do |exc|
                   code = exc.respond_to?(:code) ? exc.code : error_code_from_class(exc)
-                  error_document.add(code, "#{exc.class}: #{exc.message}")
+                  error_document.add(code.to_s, "#{exc.class}: #{exc.message}")
                 end
               end
 
--- a/lib/insights/api/common/application_controller_mixins/exception_handling.rb
+++ b/lib/insights/api/common/application_controller_mixins/exception_handling.rb
@@ -30,7 +30,7 @@ module Insights
 
           def error_code_from_class(exception)
             if ActionDispatch::ExceptionWrapper.rescue_responses.key?(exception.class.to_s)
-              ActionDispatch::ExceptionWrapper.rescue_responses[exception.class.to_s]
+              Rack::Utils.status_code(ActionDispatch::ExceptionWrapper.rescue_responses[exception.class.to_s])
             else
               DEFAULT_ERROR_CODE
             end
--- a/spec/dummy/app/controllers/api/v1x0.rb
+++ b/spec/dummy/app/controllers/api/v1x0.rb
@@ -66,6 +66,10 @@ module Api
       rescue ArgumentError
         raise SomethingHappened, "something else happened"
       end
+
+      def http_error
+        raise ActionCable::Connection::Authorization::UnauthorizedError
+      end
     end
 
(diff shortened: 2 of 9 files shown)


lib/insights/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -11,7 +11,7 @@ def self.included(other)
               errors = Insights::API::Common::ErrorDocument.new.tap do |error_document|
                 exception_list_from(exception).each do |exc|
                   code = exc.respond_to?(:code) ? exc.code : error_code_from_class(exc)
-                  error_document.add(code, "#{exc.class}: #{exc.message}")
+                  error_document.add(code.to_s, "#{exc.class}: #{exc.message}")
```
syncrou, line 14: The only real issue with this change: our kibana indexes will need to be rebuilt, most likely for all services using this gem.

That can be done, but it should be planned pretty quickly after this change is pulled into a microservice. Talking to @coderbydesign it is tough to troubleshoot why Kibana isn't returning what you expect with existing filters.

I'm also thinking filters would need to adjusted to handle the string based status code, but that is an assumption.

syncrou: Prometheus rules may be fine with this change.  It seemed that the gateway normalizes status codes.

mkanoor: @syncrou Not sure how this would cause problems in Kibana it is just a message string, so if the status started as a string or an integer it should not matter. Its just about conformance to the IPP in the payload that is returned. The JSON-API uses status as string.

PR #180, opened 2020-04-14
Percolate errors from different micro services

mkanoor:
https://projects.engineering.redhat.com/browse/SSP-1437

Since Catalog interacts with multiple micro services, RBAC, Approval,
Topology etc, we need to percolate errors from the other services upto
the caller. Since the IPP recommends using JSON-API's error format which
is an array of error objects. This PR handles errors coming from
openapi-generated clients and converts them into error objects.
Screen Shot 2020-04-14 at 6 15 06 PM

### DIFF
--- a/lib/insights/api/common/application_controller_mixins/exception_handling.rb
+++ b/lib/insights/api/common/application_controller_mixins/exception_handling.rb
@@ -46,7 +46,7 @@ module Insights
 
           def api_client_errors(exc, error_document)
             body = JSON.parse(exc.response_body)
-            if body.key?('errors') && body['errors'].is_a?(Array)
+            if body.is_a?(Hash) && body.key?('errors') && body['errors'].is_a?(Array)
               body['errors'].each do |error|
                 next unless error.key?('status') && error.key?('detail')
 
--- a/lib/insights/api/common/application_controller_mixins/exception_handling.rb
+++ b/lib/insights/api/common/application_controller_mixins/exception_handling.rb
@@ -40,7 +40,8 @@ module Insights
           end
 
           def api_client_exception?(exc)
-            exc.respond_to?(:code) && exc.respond_to?(:response_body) && exc.respond_to?(:response_headers)
+            exc.respond_to?(:code) && exc.respond_to?(:response_body) && exc.respond_to?(:response_headers) &&
+              !exc.response_body.nil?
           end
 
           def api_client_errors(exc, error_document)
--- a/lib/insights/api/common/application_controller_mixins/exception_handling.rb
+++ b/lib/insights/api/common/application_controller_mixins/exception_handling.rb
@@ -53,7 +53,7 @@ module Insights
                 error_document.add(error['status'], error['detail'])
               end
             else
-              error_document.add(exc.code, exc.message )
+              error_document.add(exc.code.to_s, exc.message )
             end
           end
         end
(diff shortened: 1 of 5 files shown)


lib/insights/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -35,6 +38,24 @@ def error_code_from_class(exception)
               DEFAULT_ERROR_CODE
             end
           end
+
+          def api_client_exception?(exc)
+            exc.respond_to?(:code) && exc.respond_to?(:response_body) && exc.respond_to?(:response_headers) &&
+              !exc.response_body.nil?
```
hsong-rh, line 44: Should `!exc.response_body.nil?` check be here? ApiException itself allows response_body to be nil. Better move it into method `api_client_errors`

lib/insights/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -35,6 +38,24 @@ def error_code_from_class(exception)
               DEFAULT_ERROR_CODE
             end
```
hsong-rh, line 39: I know this doesn't belong to your changes. Should we change it into string?

lib/insights/api/common/application_controller_mixins/exception_handling.rb
mkanoor: @hsong-rh I was doing that initially in the api_client_errors the reason I moved it here is because if the body is nil then we want to treat it as a regular exception and add a single error object.

lib/insights/api/common/application_controller_mixins/exception_handling.rb
mkanoor: @hsong-rh This still needs to be an integer because this is the HTTP status value which is always an integer, when we create our errors array we set the status and convert the into to string to comply with the JSON-API spec.

hsong-rh: Yes, I found the conversion `error_code_from_class(exc).to_s`

lib/insights/api/common/application_controller_mixins/exception_handling.rb
syncrou, line 44: `exc.respond_to?(:code) && exc.respond_to?(:response_body) && exc.respond_to?(:response_headers) &&
              !exc.response_body.nil?` Is there no other way than parsing all the `attr_reader`s to glean that this is the exception you're looking for?

lindgrenj6: I was thinking about this as well - maybe we could have it just check the name of the class, for example `RBACApiClient::ApiError`/`ApprovalApiClient::ApiError` both would match `/Client::ApiError/`. Maybe we could have it do a string match on it? or would that not be close enough?

mkanoor: @syncrou @lindgrenj6 
This exception class is a generic format for openapi_generated clients for examples for the 3 different clients we use 
Approval
https://github.com/RedHatInsights/approval-api-client-ruby/blob/5bdb4c7fecc44693ada76f93b8cd60d1a7c8f241/lib/approval-api-client-ruby/api_error.rb#L14
Topology
https://github.com/RedHatInsights/topological_inventory-api-client-ruby/blob/8c1448c7883b4f7f1c13660447a61025fa8ae5b0/lib/topological_inventory-api-client/api_error.rb#L15
RBAC
https://github.com/RedHatInsights/insights-rbac-api-client-ruby/blob/3d42dc6f6485627e6bbeb0955285b8f72a4a02d2/lib/rbac-api-client/api_error.rb#L15

Plus there is no guarantee that there would be a response body. When the API Client cannot establish a network connection we don't get back a response body. So checking for those 3 methods allows us to figure out that we have a valid api error object.

PR #182, opened 2020-04-30
Enforce unique operation ids

mkanoor:
If there are duplicate operationIds we raise an exception with the generated method name and the offending class.
```
ArgumentError: operation id cannot be duplicates, listApplicationAuthentications in class ApplicationAuthentication has already been used in class Authentication
```

This PR allows a class to override the operation Id for list operations only for now by creating a class method called list_operation_id
e.g.
```
def self.list_operation_id
  "listAllApplicationAuthentications"
end
```

This helps when we have 2 REST calls which use the same generated operationId a way to override ride as well as check for duplicates during the generations

GET /application_authentications         => listAllApplicationAuthentications
GET /applications/{id}/authentications => listApplicationAuthentications

### DIFF
--- a/lib/insights/api/common/open_api/generator.rb
+++ b/lib/insights/api/common/open_api/generator.rb
@@ -48,7 +48,7 @@ module Insights
             ENV['APP_NAME'] = app_name
             ENV['PATH_PREFIX'] = app_prefix
             Rails.application.reload_routes!
-            @operation_ids = []
+            @operation_id_hash = {}
           end
 
           def server_base_path
@@ -199,16 +199,9 @@ module Insights
 
           def openapi_list_description(klass_name, primary_collection)
             sub_collection = (primary_collection != klass_name)
-            klass = klass_name.constantize
-            operationId = if klass.respond_to?(:list_operation_id)
-                            klass.send(:list_operation_id)
-                          else
-                            "list#{primary_collection if sub_collection}#{klass_name.pluralize}"
-                          end
-            validate_operation_id(operationId, klass_name)
             {
               "summary"     => "List #{klass_name.pluralize}#{" for #{primary_collection}" if sub_collection}",
-              "operationId" => operationId,
+              "operationId" => operation_id(klass_name, primary_collection, sub_collection),
               "description" => "Returns an array of #{klass_name} objects",
               "parameters"  => [
                 { "$ref" => "##{PARAMETERS_PATH}/QueryLimit"  },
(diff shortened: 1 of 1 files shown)


lib/insights/api/common/open_api/generator.rb
```diff
@@ -562,6 +564,16 @@ def handle_custom_route_action(_route_action, _verb, _primary_collection)
           def schema_overrides
             {}
           end
+
+          def unique_name(operation_id)
+            if @operation_ids.include?(operation_id)
+              # Method names would be aa, aa_1, aa_2 so we dont mangle the first aa
+              counter = @operation_ids.select { |name| name.start_with?(operation_id) }.count
```
gmcculloug, line 571: @mkanoor Does this run in console, like a rake task?  Does it make sense to report that we are handling this conflict resolution to stdout instead of silently handling it?

mkanoor:
@gmcculloug Please review

lib/insights/api/common/open_api/generator.rb
mkanoor: @gmcculloug I have posted an example of the exception that is raised.

lib/insights/api/common/open_api/generator.rb
```diff
@@ -198,9 +199,16 @@ def openapi_schema(klass_name)
 
           def openapi_list_description(klass_name, primary_collection)
             sub_collection = (primary_collection != klass_name)
+            klass = klass_name.constantize
+            operationId = if klass.respond_to?(:list_operation_id)
+                            klass.send(:list_operation_id)
+                          else
+                            "list#{primary_collection if sub_collection}#{klass_name.pluralize}"
+                          end
+            validate_operation_id(operationId, klass_name)
```
gmcculloug, line 208: I would like to see this block of code move into an `operation_id` method that takes parameters (`klass_name, primary_collection, sub_collection`) and can just be called from line 211 below.

lib/insights/api/common/open_api/generator.rb
```diff
@@ -562,6 +570,13 @@ def handle_custom_route_action(_route_action, _verb, _primary_collection)
           def schema_overrides
             {}
           end
+
+          def validate_operation_id(operation_id, klass_name)
+            if @operation_ids.include?(operation_id)
```
gmcculloug, line 575: @mkanoor Would we want to retain the operation_id and klass_name from the conflicting element?  Maybe instead of using an array we should use a hash to store the `operation_id` as the key and the `klass_name` as the value.

The check could become:
`@operation_ids.has_key?(operation_id)` 

Or we could just do the lookup and if the result is non-`nil` we report the `operation_id` and the previous and current `klass_name`.

```ruby
def validate_operation_id(operation_id, klass_name)
  previous_klass = @operation_ids[operation_id]
  if previous_klass
    raise ArgumentError, "operation id cannot be duplicates, #{operation_id} in class #{klass_name}  Previous class #{previous_klass}"
  end
  @operation_ids[operation_id] = klass_name
end

```

mkanoor: @gmcculloug Please review

PR #184, opened 2020-05-05
Add support for custom prometheus metrics defined by the application

lindgrenj6:
This is for https://github.com/RedHatInsights/topological_inventory-ingress_api/pull/87 and potentially other applications that are currently using the Common Metrics classes. 

Basically it allows passing in a hash defined by the application to register prometheus metrics and make helper methods for them as specified.

First pass only supports counters, but maybe I'll implement support for histograms etc in the future. 

example implementation: https://github.com/RedHatInsights/topological_inventory-ingress_api/pull/87

\# TODO:
- [x] specs

### DIFF
--- a/lib/insights/api/common/metrics.rb
+++ b/lib/insights/api/common/metrics.rb
@@ -2,13 +2,14 @@ module Insights
   module API
     module Common
       class Metrics
-        def self.activate(config, prefix)
+        def self.activate(config, prefix, args = {})
           require 'prometheus_exporter'
           require 'prometheus_exporter/client'
 
           ensure_exporter_server
           enable_in_process_metrics
           enable_web_server_metrics(prefix)
+          setup_custom_metrics(args[:custom_metrics])
         end
 
         private_class_method def self.ensure_exporter_server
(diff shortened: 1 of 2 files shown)


lib/insights/api/common/metrics.rb
```diff
@@ -33,6 +34,23 @@ def self.activate(config, prefix)
           require "insights/api/common/middleware/web_server_metrics"
           Rails.application.middleware.unshift(Insights::API::Common::Middleware::WebServerMetrics, :metrics_prefix => prefix)
         end
+
+        private_class_method def self.setup_custom_metrics(custom_metrics)
+          return if custom_metrics.nil?
+
+          custom_metrics.each do |metric|
+            instance_variable_set("@#{metric[:name]}_#{metric[:type]}", PrometheusExporter::Client.default.register(metric[:type], metric[:name], metric[:description]))
```
syncrou, line 42: Is there any way to validate or map a list of values that `:name, :type, :description` can be? Potentially like `constantize_with_care`

In this current form this seems like we'll `eval` any of the code keys coming in and that seems like potential for a problem.

lib/insights/api/common/metrics.rb
```diff
@@ -33,6 +34,23 @@ def self.activate(config, prefix)
...
+
+            define_singleton_method(metric[:method]) do
+              case metric[:type]
+              when :counter
```
syncrou, line 46: I do see that `:type` can only be `:counter` which makes me feel a lot better.

lindgrenj6:
@syncrou I updated the implementation to only require `:name`, `:type`, and `:description`. This way it'll just define methods based on the name, the instance_var is just the name parameter with `_#{type}` appended since it's not really exposed by the api. Let me know what you think!

syncrou:
@lindgrenj6 - Looks good but seems to be some Travis issues.

syncrou:
@mkanoor would you mind taking a look?

PR #185, opened 2020-05-06
Handle Pundit::NotAuthorizedError exceptions differently than general exceptions

eclarizio:
In Catalog, we have an error that is being raised by a gem automatically, `Pundit::NotAuthorizedError`, which also auto-generates a not-so-user-friendly error message of `"not authorized to <action?> this Class"`.

We needed a way to hook into the `rescue_from` process without duplicating the code in common, so I broke out the handler into a separate method and I'm using it to pass control over the error message to the `rescue_from` we have for specifically handling `Pundit::NotAuthorizedError`s.

Example usage: [Catalog PR](https://github.com/RedHatInsights/catalog-api/pull/731).

Edit: I still broke out the handling code into a `rescue_from_handler` since it was a simple refactor and it allows us to potentially use it going forward, but we are not currently using it. Instead, common is just going to attempt to handle any Pundit::NotAuthorizedError exception similar to how it branches when handling api client errors.

### DIFF
--- a/lib/insights/api/common/application_controller_mixins/exception_handling.rb
+++ b/lib/insights/api/common/application_controller_mixins/exception_handling.rb
@@ -7,19 +7,25 @@ module Insights
 
           def self.included(other)
             other.rescue_from(StandardError, RuntimeError) do |exception|
-              logger.error("#{exception.class.name}: #{exception.message}\n#{exception.backtrace.join("\n")}")
-              errors = Insights::API::Common::ErrorDocument.new.tap do |error_document|
-                exception_list_from(exception).each do |exc|
-                  if api_client_exception?(exc)
-                    api_client_errors(exc, error_document)
-                  else
-                    error_document.add(error_code_from_class(exc).to_s, "#{exc.class}: #{exc.message}")
-                  end
+              rescue_from_handler(exception) do |error_document, exc|
+                error_document.add(error_code_from_class(exc).to_s, "#{exc.class}: #{exc.message}")
+              end
+            end
+          end
+
+          def rescue_from_handler(exception)
+            logger.error("#{exception.class.name}: #{exception.message}\n#{exception.backtrace.join("\n")}")
+            errors = Insights::API::Common::ErrorDocument.new.tap do |error_document|
+              exception_list_from(exception).each do |exc|
+                if api_client_exception?(exc)
+                  api_client_errors(exc, error_document)
+                else
+                  yield error_document, exc
                 end
               end
-
-              render :json => errors.to_h, :status => error_code_from_class(exception)
             end
+
+            render :json => errors.to_h, :status => error_code_from_class(exception)
           end
 
           def exception_list_from(exception)
--- a/lib/insights/api/common/application_controller_mixins/exception_handling.rb
+++ b/lib/insights/api/common/application_controller_mixins/exception_handling.rb
@@ -4,7 +4,6 @@ module Insights
       module ApplicationControllerMixins
         module ExceptionHandling
           DEFAULT_ERROR_CODE = 400
-          PUNDIT_UNAUTHORIZED_ERROR = "Pundit::NotAuthorizedError".freeze
 
           def self.included(other)
             other.rescue_from(StandardError, RuntimeError) do |exception|
@@ -20,8 +19,8 @@ module Insights
               exception_list_from(exception).each do |exc|
                 if api_client_exception?(exc)
                   api_client_errors(exc, error_document)
-                elsif pundit_exception?(exc)
-                  message = "You are not authorized to #{exc.query.delete_suffix('?')} this #{exc.record.model_name.human.downcase}"
+                elsif custom_exception?(exc)
+                  message = fetch_custom_message(exc)
                   error_document.add(error_code_from_class(exc).to_s, message)
                 else
                   yield error_document, exc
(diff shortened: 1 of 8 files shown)


lib/insights/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -7,19 +7,25 @@ module ExceptionHandling
...
-              errors = Insights::API::Common::ErrorDocument.new.tap do |error_document|
-                exception_list_from(exception).each do |exc|
-                  if api_client_exception?(exc)
-                    api_client_errors(exc, error_document)
-                  else
-                    error_document.add(error_code_from_class(exc).to_s, "#{exc.class}: #{exc.message}")
-                  end
-                end
+              rescue_from_handler(exception) do |error_document, exc|
+                error_document.add(error_code_from_class(exc).to_s, "#{exc.class}: #{exc.message}")
```
mkanoor, line 11: @eclarizio is the exc.class too much information to the user? Should we just have the message.
Also now that we have the JSON-API errors array, the HTTP status code might get handled by the UI and it might display Authorization error for the 403 and the errors array will have the details.

https://projects.engineering.redhat.com/browse/RHCLOUD-5966

lib/insights/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -7,19 +7,25 @@ module ExceptionHandling
...
               end
+            end
+          end
 
-              render :json => errors.to_h, :status => error_code_from_class(exception)
+          def rescue_from_handler(exception)
+            logger.error("#{exception.class.name}: #{exception.message}\n#{exception.backtrace.join("\n")}")
+            errors = Insights::API::Common::ErrorDocument.new.tap do |error_document|
+              exception_list_from(exception).each do |exc|
+                if api_client_exception?(exc)
```
mkanoor, line 20: @eclarizio can we check if the class name is PunditAuthorizationException and tweak the message here like we do with api exceptions from generated clients. This way it would help all other users of Pundit.

lib/insights/api/common/application_controller_mixins/exception_handling.rb
eclarizio: @mkanoor Possibly `exc.class` is too much information for the user, but this is just a refactor, I didn't change any logic. I think if we want to change the functionality here, that should be a separate PR. This is simply to extract the handling code into a non-anonymous method so that it is usable.

lib/insights/api/common/application_controller_mixins/exception_handling.rb
eclarizio: I suppose we could, but doesn't that then make this gem sort of depend upon pundit just for the exception declaration? I don't know if I like that idea.

lindgrenj6: I'm ok with checking the class name there, we do it for all of the generated clients already so another check fits in line. Both Catalog and Approval are using pundit so far right?

eclarizio:
@syncrou @lindgrenj6 It looks like travis is failing because it's using an older ruby version that doesn't have the `#delete_suffix` method for strings. Should we be updating ruby version dependency for this repo or should I use the older `.chomp('?')`? Seems like 2.4.5 is a bit old, but I don't know if 2.5.x breaks anything else we use in this repo.

gmcculloug:
@eclarizio @lindgrenj6 My understanding is all the apps are running on 2.5.5 and travis for a number of repos is testing against [2.5 and 2.6](https://github.com/RedHatInsights/sources-api/blob/master/.travis.yml#L5-L7).  

Seems we should bump the travis rvm version here.

lindgrenj6:
@gmcculloug @eclarizio https://github.com/RedHatInsights/insights-api-common-rails/pull/187

Opened a PR for here and catalog/approval to keep us all in line.

gmcculloug:
@eclarizio ruby version PR is merged.  Please rebase.

eclarizio:
@gmcculloug @lindgrenj6 Rebased/updated. Looks like we're green now! 👍

lib/insights/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -4,22 +4,32 @@ module Common
...
+                  api_client_errors(exc, error_document)
+                elsif pundit_exception?(exc)
```
syncrou, line 23: It feels odd to have such a specific exception class in this generic class. More so because before this PR Pundit was handled in only the ApprovalApi / CatalogAPI apps.

lib/insights/api/common/application_controller_mixins/exception_handling.rb
```diff
@@ -4,22 +4,32 @@ module Common
       module ApplicationControllerMixins
         module ExceptionHandling
           DEFAULT_ERROR_CODE = 400
+          PUNDIT_UNAUTHORIZED_ERROR = "Pundit::NotAuthorizedError".freeze
```
syncrou, line 7: What about a `pundit_exception_handling` class to define these custom constants in.

lib/insights/api/common/application_controller_mixins/exception_handling.rb
eclarizio: I wasn't really being handled in those apps, it was just falling through to the last `else` here, and the message that that case spits out was not very user friendly.

eclarizio:
@syncrou Pushed changes, is this more what you're looking for?

PR #188, opened 2020-05-07
The status field in the error object is string

mkanoor:
To conform with the IPP and the JSON API error object format
the generator should be using string for status.

This PR updates the pre-canned type used by the generator used by topology and sources when they generate the openapi.json so it can create the JSON API conformant type.

https://github.com/RedHatInsights/sources-api/blob/8782d2103bed61c40b547603089c0978624950e1/public/doc/openapi-3-v3.0.json#L1780

### DIFF
--- a/lib/insights/api/common/open_api/generator.rb
+++ b/lib/insights/api/common/open_api/generator.rb
@@ -117,8 +117,8 @@ module Insights
                     "type"        => "object",
                     "properties"  => {
                       "status"    => {
-                        "type"    => "integer",
-                        "example" => 404
+                        "type"    => "string",
+                        "example" => "404"
                       },
                       "detail"    => {
                         "type"    => "string",


mkanoor:
@eclarizio The openapi generator doesn't have any specs, It runs thru app/models and generates an openapi.json for sources and topology. The changes in this PR is part of a precanned section that is inserted into the openapi.json file. This is only for documentation purposes when an error is sent out from the api-server we create an ErrorDcoument that conforms to this spec. It becomes part of the HTTP response body and at that point it doesn't matter if its a string or an integer.

We do have failures when we test the error response from a api request. They are handled separately in this PR
https://github.com/RedHatInsights/sources-api/pull/219

There are tons of failure that is being fixed in that PR.

This PR just fixes the documentation aspect of it so that we dont keep inserting the precanned object with integer status.

PR #196, opened 2020-05-28
Add the ability to disable metrics by setting METRICS_PORT=0

lindgrenj6:
Found out that we can't really disable these metrics, so we run into issues with multiple common-enabled apps that can't bind to the same port if running on localhost. 

This allows for disabling or setting a different port number if you really want the metrics.

### DIFF
--- a/lib/insights/api/common/metrics.rb
+++ b/lib/insights/api/common/metrics.rb
@@ -6,6 +6,8 @@ module Insights
           require 'prometheus_exporter'
           require 'prometheus_exporter/client'
 
+          return if metrics_port == 0
+
           ensure_exporter_server
           enable_in_process_metrics
           enable_web_server_metrics(prefix)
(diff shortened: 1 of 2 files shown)


lib/insights/api/common/metrics.rb
```diff
@@ -51,6 +53,10 @@ def self.activate(config, prefix, args = {})
             end
           end
         end
+
+        private_class_method def self.metrics_port
+          @metrics_port ||= (ENV['METRICS_PORT']&.to_i || 9394)
```
gmcculloug, line 58: @lindgrenj6 Change looks good, just wanted to ask if we feel comfortable with the ENV var name.  It is short and generic so not sure if we would run into an issue anywhere.

My only other name suggestion would be `INSIGHTS_METRICS_PORT`.  But if others don't see an issue (because it is only for local test envs) then I think it is fine.

lindgrenj6: The main reason for the name is because all of the other services that export prometheus metrics use the same `ENV` var, I'd be down for changing it to something else like `INSIGHTS_METRIC_PORT`/`PROMETHEUS_PORT` since that's even more descriptive, but I think I'd want to change it everywhere to be consistent.

gmcculloug: Good to know.  Makes sense to keep the current value to be consistent.  👍

PR #198, opened 2020-06-05
Excluding existing 'nullable' attributes

slemrmartin:
`nullable` attributes need to be kept when generating new json

**related**: https://github.com/RedHatInsights/sources-api/pull/231

### DIFF
--- a/lib/insights/api/common/open_api/generator.rb
+++ b/lib/insights/api/common/open_api/generator.rb
@@ -441,7 +441,7 @@ module Insights
               end
 
               # Take existing attrs, that we won't generate
-              ['example', 'format', 'readOnly', 'title', 'description'].each do |property_key|
+              ['example', 'format', 'nullable', 'readOnly', 'title', 'description'].each do |property_key|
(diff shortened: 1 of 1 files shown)


lib/insights/api/common/open_api/generator.rb
```diff
@@ -441,7 +441,7 @@ def openapi_schema_properties_value(klass_name, model, key, value)
               end
 
               # Take existing attrs, that we won't generate
-              ['example', 'format', 'readOnly', 'title', 'description'].each do |property_key|
+              ['example', 'format', 'nullable', 'readOnly', 'title', 'description'].each do |property_key|
```
gmcculloug, line 444: Looks good, let's add a spec to cover this.

slemrmartin: @gmcculloug there are no specs for whole generator so it'll be quite hard to write specs for one attr

syncrou: JIRA Ticket: https://projects.engineering.redhat.com/browse/TPINVTRY-977

PR #199, opened 2020-06-17
Add github workflow to release to rubygems automatically

lindgrenj6:
https://github.com/marketplace/actions/publish-to-rubygems

Uses that action, which just runs `rake release` at the project root, we just need two secrets:
- `GITHUB_TOKEN`
- `RUBYGEMS_API_KEY`

cc @gmcculloug @syncrou @slemrmartin

### DIFF
--- a/.github/workflows/gem-push.yml
+++ b/.github/workflows/gem-push.yml
@@ -1,0 +1,50 @@
+name: Release Ruby Gem
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+    - 'lib/insights/api/common/version.rb'
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


lindgrenj6:
@syncrou this is ready to go now that we have it working for topological_inventory-providers-common

slemrmartin:
@lindgrenj6 where are the secrets stored?

lindgrenj6:
It is in the repository settings - but I think you have to be owner to add them.

We also found out that `GITHUB_TOKEN` is implicitly added by the workflow run, so we only need to provide the rubygem key to get the workflow working.

Issue #201, opened 2020-06-26
Sharing the `openapi-validator` script from this common repo


gmcculloug:
Currently the `openapi-validator.sh` script lives in the `.travis_scripts` directory in 4 repos:
- https://github.com/RedHatInsights/approval-api
- https://github.com/RedHatInsights/catalog-api
- https://github.com/RedHatInsights/sources-api
- https://github.com/RedHatInsights/topological_inventory-api

While this file is not expected to change much, when it does it requires updating all 4 repos.  Since the script is designed to run during travis tests we can include it into this common repo and either 
- pull the raw file directly during from the travis script
- reference the script based on the locally installed bundle path: `bundle info insights-api-common --path`

PR #202, opened 2020-06-27
Add openapi-validator.sh script for reference from API repos.

gmcculloug:
Part of fixing #201

Move the `openapi-validator.sh` script to a common location for all the API repos to access.

### DIFF
--- a/.travis_scripts/openapi-validator.sh
+++ b/.travis_scripts/openapi-validator.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+set -eu
+set -o pipefail
+
+if ! ver=$(./openapi-generator-cli version | tail -1); then
+  ver="4.3.1"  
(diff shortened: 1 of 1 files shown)


lindgrenj6:
I think we need to add `.travis_scripts/` to the gemspec here:
https://github.com/RedHatInsights/insights-api-common-rails/blob/bc964c7ec61028e311215b5df29d368c2ca30333/insights-api-common.gemspec#L14

Looking at it now it doesn't look like it's included. Of course this all depends if we decide to download it from github or use the version bundled with the gem.

gmcculloug:
@lindgrenj6 Good point.  I did end up switching to downloading the script from the `master` branch so we probably can avoid adding it to the gem unless others thing it should be there for some reason.  For now, I'll leave the PR as-is and can update if there is more feedback on this topic.

PR #204, opened 2020-07-10
Add download of openapi-generator script

gmcculloug:
After spending some time with other travis scripts I started thinking it would be better to move the download of `openapi-generator-cli` to this common method instead of requiring each caller to have to download it separately as part of the `.travis.yaml`

It's not a big impact here (2 new lines), but the caller's `.travis.yml` file can be reduced to one line:
```diff
- - curl -L https://raw.githubusercontent.com/OpenAPITools/openapi-generator/master/bin/utils/openapi-generator-cli.sh
-   > ./openapi-generator-cli
- - chmod +x ./openapi-generator-cli
- - curl -L https://raw.githubusercontent.com/RedHatInsights/insights-api-common-rails/master/.travis_scripts/openapi-validator.sh
-   > ./openapi-validator.sh
- - chmod +x ./openapi-validator.sh
- - ./openapi-validator.sh
+ - curl -sSL https://raw.githubusercontent.com/gmcculloug/insights-api-common-rails/master/.travis_scripts/openapi-validator.sh | bash -s
```

If you like this approach it can be merged ahead of updating the callers to use the 1-line call.

### DIFF
--- a/.travis_scripts/openapi-validator.sh
+++ b/.travis_scripts/openapi-validator.sh
@@ -2,6 +2,10 @@
 set -eu
 set -o pipefail
 
+curl -L https://raw.githubusercontent.com/OpenAPITools/openapi-generator/master/bin/utils/openapi-generator-cli.sh > ./openapi-generator-cli
+chmod +x ./openapi-generator-cli
+echo "openapi-generator-cli successfully downloaded"
+
 if ! ver=$(./openapi-generator-cli version | tail -1); then
   ver="4.3.1"  
   echo "Use default version ${ver} to validate"


.travis_scripts/openapi-validator.sh
```diff
@@ -2,6 +2,9 @@
 set -eu
 set -o pipefail
 
+curl -L https://raw.githubusercontent.com/OpenAPITools/openapi-generator/master/bin/utils/openapi-generator-cli.sh > ./openapi-generator-cli
+chmod +x ./openapi-generator-cli
+
```
bzwei, line 7: Shall we add an `echo` here to indicate the curl download is completed?

gmcculloug: @bzwei Added an `echo`.  Is that what you were looking for?

lindgrenj6:
@gmcculloug does this need a new gem release? or is this script just downloaded from master in the child-repos?

gmcculloug:
@lindgrenj6 No, the callers pull this file directly from the master branch during the travis test run.

Example: https://github.com/RedHatInsights/sources-api/pull/244/files

PR #205, opened 2020-07-27
Make sure Pundit exception's query type is String

hsong-rh:
This PR allows `authorize` to accept symbol as its second parameter, like `authorized(Portfolio, :update?)`, otherwise following `delete_suffix` will fail and raise other exceptions.


spec/lib/insights/api/common/custom_exceptions_spec.rb
```diff
@@ -10,5 +11,15 @@
         )
       end
     end
+
+    context "when the query is Symbol" do
+      let(:exception) { double(:class => "Pundit::NotAuthorizedError", :query => :create?, :record => record) }
```
eclarizio, line 16: Small nitpick, but what do you think about moving this below line 3, changing to use `:query => query` and then use `let(:query) { :create? }` for this context and `let(:query) { "create?" }` for the other context?

PR #208, opened 2020-09-22
Fixed to log error messages in with_request block

hsong-rh:
When exception is raised, the place to log error messages has lost the request context. This PR moves the log call within request context.

### DIFF
--- a/spec/lib/insights/api/common/request_spec.rb
+++ b/spec/lib/insights/api/common/request_spec.rb
@@ -200,6 +200,8 @@ describe Insights::API::Common::Request do
     end
 
     it 'with an invalid Hash' do
+      expect(Rails.logger).to receive(:error).with(/ArgumentError: /).once
+
       expect do
         described_class.with_request({}) {}
       end.to raise_exception(ArgumentError)
--- a/lib/insights/api/common/application_controller_mixins/exception_handling.rb
+++ b/lib/insights/api/common/application_controller_mixins/exception_handling.rb
@@ -14,7 +14,6 @@ module Insights
           end
 
           def rescue_from_handler(exception)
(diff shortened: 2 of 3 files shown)


hsong-rh:
@lindgrenj6 @mkanoor Please review

eclarizio (commented):
LGTM. Not sure if there's a way to add specs to this, since it's kind of rearranging the error flow and that feels difficult to encapsulate in a spec, but that would be great if we could.

mkanoor (approved):

spec/lib/insights/api/common/request_spec.rb
```diff
@@ -200,6 +200,8 @@
     end
 
     it 'with an invalid Hash' do
+      expect(Rails.logger).to receive(:error).with(/ArgumentError: /).once
```
eclarizio, line 203: The point of this PR was to "keep the request_id" in the logs, but if we're not able to show it here, I don't know if this accomplishes much.

hsong-rh: This test is only to make sure error message is logged, not to prove `request_id` is kept. The underlying cloudwatch logger will put `request_id` in Kibana log. I renamed the confused topic as: Fixed to log error messages in with_request block.

mkanoor: @eclarizio This request_id is logged by manageiq logger, it checks the Thread Local Storage to see if there is a request.
That code is here
https://github.com/ManageIQ/manageiq-loggers/blob/241ebecd7930a481733fcbea93ebde92f4226acc/lib/manageiq/loggers/container.rb#L67

eclarizio (approved):

PR #210, opened 2020-10-26
Add rubocop_base.yml for common rubocop rules

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-10237

Rather than making a new repo lets use api-common as the main repo. We will need to update ALL THE THINGS after this is merged though!

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -1,0 +1,291 @@
+require:
+  - rubocop-rails
+  - rubocop-performance
+
+AllCops:
+  NewCops: disable
+  Exclude:
+    - vendor/**/*
+  TargetRubyVersion: 2.6
+Layout/HashAlignment:
+  EnforcedHashRocketStyle: table
+  EnforcedColonStyle: table
+Layout/DefEndAlignment:
+  AutoCorrect: true
+Layout/EndAlignment:
+  AutoCorrect: true
+Layout/ExtraSpacing:
+  AutoCorrect: true
+Layout/LineLength:
+  Enabled: false
+Layout/SpaceAroundMethodCallOperator:
+  Enabled: true
+Layout/SpaceBeforeFirstArg:
+  Enabled: false
+Layout/SpaceInsideHashLiteralBraces:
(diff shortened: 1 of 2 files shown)


eclarizio:
@lindgrenj6 I'm fine with this if we want to keep it in here instead of the new "manageiq-style" repo, but I think with this change we have to update rubocop on dippy-bot to a later version. I made a similar PR for Catalog: https://github.com/RedHatInsights/catalog-api/pull/844 but dippy-bot is also complaining about missing linter config files.

.rubocop_base.yml
```diff
@@ -0,0 +1,292 @@
+require:
+  - rubocop-rails
+  - rubocop-performance
+
+AllCops:
+  NewCops: disable
+  Exclude:
+    - spec/manageiq/**/*
```
slemrmartin, line 8: this is not needed

.rubocop_base.yml
```diff
@@ -0,0 +1,292 @@
...
+    - vendor/**/*
+  TargetRubyVersion: 2.5
```
slemrmartin, line 10: can be switched to 2.6?

slemrmartin (commented):
Just a few comments, otherwise LGTM (I'm not reviewing the rules, it's not a subject of this PR)

.rubocop_base.yml
lindgrenj6: Sure, updated.

lindgrenj6:
@eclarizio yeah I think that is one thing we are going to try to do now/in the future. It would be really nice to upgrade to rubocop 1.0 since that was just released recently.

syncrou (approved):
I'm good with this. For anyone who wants to use the old MIQ rules ( at the new repo ) this change allows that. For those wishing to manage them internal to RedHatInsights, this change allows for that. LGTM 👍

syncrou:
@eclarizio @lindgrenj6 with the questions about Dippy Bot, that doesn't hold this PR up does it?

PR #214, opened 2021-01-11
Use custom error_message if a policy exposes it

eclarizio:
When we check requirements for actions, we generally check against the parent Portfolio if the user has the permission to update the Portfolio. This happens for example when creating portfolio items or when editing surveys, resetting service plans, and various other actions for portfolio items and service plans.

However, because we're checking against a Portfolio, the way Pundit handles it if the check fails is that the record stored in the exception is now `Portfolio`, which ends up generating bad messages because the user will see "You are not authorized to _action_ for this Portfolio" when they were trying to create/edit either a portfolio item or a service plan. There's not a really great way that I could find to directly override the `exception.record` without monkeypatching Pundit itself, and that felt pretty gross.

This change is very minimal to the common repo in favor of pushing the responsibility of setting a correct error message on the policy side. If the policy wishes to have a custom message, they simply set the instance variable in the method where they would like the custom message, and this should handle it. If not, we will default to the default message, although I did change the wording a little bit to better reflect the changes I made to the policy error messages.

For more info,
Please see the catalog-api PR here: https://github.com/RedHatInsights/catalog-api/pull/883
And the JIRA story here: https://issues.redhat.com/browse/SSP-1977

@lindgrenj6 Please review, or tag someone who you should think review. Thanks!

### DIFF
--- a/lib/insights/api/common/custom_exceptions.rb
+++ b/lib/insights/api/common/custom_exceptions.rb
@@ -7,7 +7,8 @@ module Insights
         def self.custom_message(exception)
           case exception.class.to_s
           when "Pundit::NotAuthorizedError"
-            "You are not authorized to #{exception.query.to_s.delete_suffix('?')} this #{exception.record.model_name.human.downcase}"
+            exception.policy.try(:error_message) ||
+              "You are not authorized to perform the #{exception.query.to_s.delete_suffix('?')} action for this #{exception.record.model_name.human.downcase}"
           end
         end
       end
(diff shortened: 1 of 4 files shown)


eclarizio:
@brahmanim Tagging you here as well Benny because once this goes through the error message language is slightly changed, so want to make sure you are aware.

lindgrenj6 (approved):
:+1: LGTM, glad to see we didn't have to monkey-patch Pundit, that would have been gnarly!

PR #215, opened 2021-01-13
Release v5.0.0

lindgrenj6:
Debating naming this release `4.2.0` instead of `4.1.6` due to the fact that `params_for_update` behavior is changing. Theoretically it should be a bug fix, but the behavior is changing nonetheless.

EDIT: After talking with Drew we decided to release 5.0 instead of a minor release due to the fact this is a breaking change that could introduce small intricacies in catalog and other apps using this code. 

cc @hsong-rh @eclarizio @syncrou

*DEPENDS ON*: 
- [x] #213
- [x] #214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [5.0.0] - 2021-01-13
+### Changed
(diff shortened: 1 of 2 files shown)


lindgrenj6:
@eclarizio We are doing a major release with #213 being in this release - I would check against that branch in catalog and see if everything is still working, and then update your Gemfile accordingly.

slemrmartin (approved):

PR #217, opened 2021-02-02
feat(logger): introduce audit logging

vkrizan:
Introduction of a middleware and utility to add entries to an audit log.  By default it logs them to STDOUT, but eventually it can be tied into CloudWatch. Log entries are formatted into JSON, split by a new-line. Usage is documented in README.

TODO:
- [X] Capture Rails requests
- [ ] Add tests
- [ ] Find a way to have a shortcut to the logger
- [ ] Support within a Sidekiq
- [ ] Support within a Racecar

Reference: RHICOMPL-1217


vkrizan:
@akofink, here is the audit logger to get an idea. It is still experimental, and if it would not be fitted in this common package we can bundle it in our Compliance app. Feel free to check it out and comment. Thanks!

lib/insights/api/common/audit_log/middleware.rb
```diff
@@ -0,0 +1,92 @@
...
+        @evidence = {}
+      end
+
+      def call(env)
+        subscribe
+        @request = ActionDispatch::Request.new(env)
+        @app.call(env).tap do |status, _headers, _body|
+          @status = status
+          response_finished
+        end
```
akofink, line 19: Nice! :100:

lib/insights/api/common/audit_log/middleware.rb
```diff
@@ -0,0 +1,92 @@
...
+        status_label = Rack::Utils::HTTP_STATUS_CODES[status]
+        msg = "#{request.method} #{request.original_fullpath} -> #{status} #{status_label}"
+        if evidence[:unpermitted_parameters]
+          msg += "; unpermitted params #{fmt_params(evidence[:unpermitted_parameters])}"
+        end
+        if evidence[:halted_callback].present?
+          msg += "; filter chain halted by :#{evidence[:halted_callback]}"
+        end
+        msg
+      end
```
akofink, line 45: Are the headers logged anywhere? It may be good to include, but it may also present a security risk to include the X_RH_IDENTITY header.

lib/insights/api/common/audit_log/middleware.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      def fmt_controller(payload)
+        return if payload[:controller].blank?
+
+        [payload[:controller], payload[:action]].compact.join('#')
+      end
+
+      def fmt_params(params)
+        params.map { |e| ":#{e}" }.join(", ")
+      end
```
akofink, line 89: Does it work with rails parameter filtering for sensitive params? We should verify this/add a test for this case.

https://guides.rubyonrails.org/action_controller_overview.html#log-filtering

akofink (commented):
Nice work, @vkrizan! Just a few minor comments/questions, but it seems reasonable to me. I'd like to revisit it to actually test it with compliance once it's ready.

lib/insights/api/common/audit_log/middleware.rb
vkrizan: Inspecting `X_RH_IDENTITY` is a good idea, and can be done here. That's mainly the reason why this is a middleware (to access the request and headers).

lib/insights/api/common/audit_log/middleware.rb
vkrizan: This only lists the params, not their value. The code snippet is strait from Rails logger.

akofink: Ah! I missed that. Is it possible to log the param values? I think it'd be really useful to be able to debug errors/malfunctions with requests, but we need to ensure they're filtered based on the built in log filtering.

lib/insights/api/common/audit_log/middleware.rb
akofink: The headers may also include `X_RH_REQUEST_ID` (or something similar), which can be important to tracing uploads, etc. across services.

lib/insights/api/common/audit_log/formatter.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      private
+
+      def format_datetime(time)
+        time.utc.strftime('%Y-%m-%dT%H:%M:%S.%6NZ')
+      end
+
+      def account_number
+        Thread.current[:audit_account_number]
+      end
```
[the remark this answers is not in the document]
vkrizan: There is only a one instance of Formatter. Setting it on that instance would not work correctly on multi-threaded env.

README.md
```diff
@@ -25,7 +25,36 @@ Or install it yourself as:
...
+Setting context of an account (e.g. an authenticated account) can be done with these options
+```
+Insights::API::Common::AuditLog.with_account('12345')
+
+Insights::API::Common::AuditLog.with_account('12345') do
+  # limited context
+end
+```
+
+### Insights::Api::Common::Filter
```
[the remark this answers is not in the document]
vkrizan: This one I was not sure about. I'd let owners to look into it. The "Usage" is H2, so the next one should be H3. I'm not sure whether the "Sorting Results" are part of the Filter or not.

lib/insights/api/common/audit_log.rb
```diff
@@ -0,0 +1,41 @@
...
+        self.logger = logger
+      end
+
+      def with_account(account_number)
+        original = Thread.current[:audit_account_number]
+        Thread.current[:audit_account_number] = account_number
+        return unless block_given?
+
+        yield
+        Thread.current[:audit_account_number] = original
```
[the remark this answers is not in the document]
vkrizan: We use hooks for the auth, which means that the `account_number` would be only setting. Ensuring that the account num is reset is one thing I'm working on. It would not be a 100%, but at least it should cover controllers.

lib/insights/api/common/audit_log/middleware.rb
vkrizan: It would be possible, but the notification does not provide them. https://guides.rubyonrails.org/active_support_instrumentation.html#unpermitted-parameters-action-controller

Please keep in mind that this is a audit log with a purpose of retracing steps one has taken. Mostly access/denial/changes to a resource, etc. If details would be necessary an app log can be of help. Hence the `transation_id`s. :wink:

lib/insights/api/common/audit_log/middleware.rb
vkrizan: Awesome! I guess we can incorporate that in the second iteration.

README.md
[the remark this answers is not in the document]
akofink: Nothing says you have to use headers in order H1 -> H2 -> H3 etc. H1 -> H3 -> H5 works just as well, probably chosen so the markdown renders more nicely in the GitHub UI without overly large headers.

lib/insights/api/common/audit_log/formatter.rb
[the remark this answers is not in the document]
akofink: There will be one instance of Formatter per thread as written afaik. Thread.current represents only the local thread, so it also won't share information across threads in a multi-threaded env.

README.md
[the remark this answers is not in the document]
vkrizan: There are document composition guidelines, hence the structure (and order).

lib/insights/api/common/audit_log/formatter.rb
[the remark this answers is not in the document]
vkrizan: > Thread.current represents only the local thread, so it also won't share information across threads in a multi-threaded env.

^ That's the sole purpose of this.

I'd like to keep it defensive as is. I don't think that there is a separate instance for each thread as it is instantiated once with a logger. I don't want to rely on that assumption either.

lib/insights/api/common/audit_log.rb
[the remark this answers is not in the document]
slemrmartin: I suggest setting the original account number in `ensure` block in case of exception

lib/insights/api/common/audit_log.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      private
+
+      def init_logger(logger = nil)
+        logger ||= Logger.new($stdout)
+        logger.level = Logger::INFO
```
slemrmartin, line 35: in topological/sources we are using ENV['LOG_LEVEL'] for ruby and ENV['RAILS_LOG_LEVEL'] for rails. 
what about `logger.level= ENV[...] || Logger::INFO`

slemrmartin (commented):
@vkrizan is the audit log intended only for APIs or also for all other services?

lib/insights/api/common/audit_log.rb
vkrizan: The level is hardcoded, as this is not going to be configurable. The idea is to use info/warn/error for audit logs. The end result might not even use and/or format the level. The difference would be only whether it is a failure/success.

vkrizan:
@slemrmartin It is a global thing, not directly related to APIs. I was operating on the fact that the main module is `Insights::API::Common`. Do you have any suggestion where to move it?  Is there other common package for Insights where we can move it?

PR #218, opened 2021-02-03
chore(deps): permissive prometheus_exporter dependency

vkrizan:
allowing to have versions 0.4 up to but not inluding 1.0 starting
with version 0.4.5.


vkrizan:
@slemrmartin, @lindgrenj6, @syncrou, would you guys be able to have a look. (I was unable to set reviewers.) Thanks.

vkrizan:
Hmm, there would be more changes to deps required to alleviate the version restrictions.

slemrmartin:
@vkrizan have you tried the gem with prometheus in app-interface, is higher version compatible?
Basically how it should influence version of this gem (patch/minor)?

vkrizan:
@slemrmartin, we use prometheus_exporter 0.5.1 in Compliance in production. I can lock it up to 0.6.

lindgrenj6:
Hey @vkrizan, thanks for the PR. Looking at this it basically locks us at `0.4.z`, not up to `0.y.z`.

If we want to lock it to basically anything pre-1.0 but greater than 0.4.5 we could lock like this: `~ 0.4, >= 0.4.5`

vkrizan:
@lindgrenj6 the suggestion looks good.

There were more of these that need to be updated in similar manor to allow us to consume the lib.

PR #219, opened 2021-03-04
[WIP] Upgrade rails to 6.1.3

hsong-rh:
Upgrade rails to latest 6.1.3

### DIFF
--- a/lib/insights/api/common/version.rb
+++ b/lib/insights/api/common/version.rb
@@ -1,7 +1,7 @@
 module Insights
   module API
     module Common
-      VERSION = "5.0.1".freeze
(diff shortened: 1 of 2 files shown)


lib/insights/api/common/version.rb
```diff
@@ -1,7 +1,7 @@
 module Insights
   module API
     module Common
-      VERSION = "5.0.1".freeze
+      VERSION = "5.0.2".freeze
```
mkanoor, line 4: @hsong-rh I think this is a breaking change, it should be 6.0.0 since its using the latest version of Rails and we could have broken things.

hsong-rh:
May work later

PR #221, opened 2021-03-19
Allow to load GraphQL erb templates outside of repo

lpichler:
This allows to make GraphQL erb([model_type.erb](https://github.com/RedHatInsights/insights-api-common-rails/blob/master/lib/insights/api/common/graphql/templates/model_type.erb), [query_type.erb](https://github.com/RedHatInsights/insights-api-common-rails/blob/master/lib/insights/api/common/graphql/templates/query_type.erb), [schema.erb](https://github.com/RedHatInsights/insights-api-common-rails/blob/master/lib/insights/api/common/graphql/templates/schema.erb)) templates configurable in repo(For example in sources-api) where is insights-api-common-rails gem used and 
templates live in location  `./api/common/graphql/templates` in remote repos.

Example: So if you want override GraphQL schema for source-api, you 
need to create file in source-api repo like is this [PR](https://github.com/RedHatInsights/sources-api/pull/348).

### Links
Together with [TOPO PR](https://github.com/RedHatInsights/topological_inventory-api/pull/350) and [Sources PR](https://github.com/RedHatInsights/sources-api/pull/348) fixes: https://issues.redhat.com/browse/RHCLOUD-11179 and disables GraphQL introspection  in source/topo  APIs.

### DIFF
--- a/lib/insights/api/common/graphql/generator.rb
+++ b/lib/insights/api/common/graphql/generator.rb
@@ -20,8 +20,32 @@ module Insights
             openapi_path.split("/")[1..-1]
           end
 
+          def self.template_file_by(type, root_dir = __dir__)
+            Pathname.new(root_dir).join(File.expand_path("templates", root_dir), "#{type}.erb")
+          end
+
+          def self.root_dir
+            Rails.root
+          end
+
+          def self.app_name
+            Rails.application.class.parent.name.underscore
+          end
+
+          def self.pluggable_template_file_by(type)
+            templates_relative_path = "lib/#{app_name}/api/graphql/templates"
+            template_path = File.expand_path(templates_relative_path, root_dir)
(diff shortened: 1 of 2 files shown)


syncrou (commented):
Would we want a spec ( or maybe we have on in the API repos ) that proves introspection doesn't work when the `disable_introspection_entry_points` config value is set?

lpichler:
> Would we want a spec ( or maybe we have on in the API repos ) that proves introspection doesn't work when the `disable_introspection_entry_points` config value is set?

yes, that's right and now I added also this case to specs,
thanks @syncrou!

syncrou (approved):
LGTM 👍

lindgrenj6 (approved):
LGTM, sorry for taking so long to see this!

